Issue #29, opened 2020-02-27
Maven build


andrewazores:
See rh-jmc-team/container-jfr#128

andrewazores:
Fixed in #30

PR #30, opened 2020-02-28
Maven Build

andrewazores:
See:
- #29
- rh-jmc-team/container-jfr#128
- rh-jmc-team/container-jfr#129

This replaces Gradle with Maven.


pom.xml
```diff
@@ -0,0 +1,216 @@
...
+    </plugin>
+  </plugins>
+</build>
+
+<reporting>
+  <plugins>
+    <plugin>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-maven-plugin</artifactId>
+      <version>3.1.12.2</version>
```
jiekang: Would it be worth defining all the plugin versions in the `properties` section? I think this one is duplicated in the file already.

jiekang:
The `mvn clean install` ran fine for me and produced sane results in the .m2. Will approve once query on versions in `properties` is resolved.

pom.xml
andrewazores: Yea, that's a good idea. Any idea or preference on how these properties should be named?

jiekang: How about:
artifactId.version with instances of '-' replaced by '.'

So:
spotbugs.maven.plugin.version

Hmm on further thought, for version definitions of plugins, you only need one instance and it could be all placed in a pluginManagement section. Something like:
https://github.com/jiekang/thermostat-web-gateway/blob/master/pom.xml

Thoughts?

jiekang: Er yeah, you described it perfectly. It's a bit verbose though and makes more sense when you have multiple maven submodules using things. Either solution is okay with me, whichever you prefer.

PR #31, opened 2020-03-19
Implement TemplateService for target VM event templates

andrewazores:
Related to rh-jmc-team/container-jfr#141
Related to rh-jmc-team/container-jfr#142

These changes support rh-jmc-team/container-jfr#140 , which adds a command to list template names on the remote VM and enables an event string syntax to select these templates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,6 @@
-.gradle
 /build/
 /target/
 /bin/
-
-# Ignore Gradle GUI config
-gradle-app.setting
-
-# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
-!gradle-wrapper.jar
-
-# Cache of project
-.gradletasknamecache
-
-# # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
-# gradle/wrapper/gradle-wrapper.properties
+.classpath
+.project
+.settings/
--- a/src/main/java/com/redhat/rhjmc/containerjfr/core/templates/RemoteTemplateService.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/core/templates/RemoteTemplateService.java
@@ -1,0 +1,98 @@
+package com.redhat.rhjmc.containerjfr.core.templates;
+
+import java.io.IOException;
+import java.text.ParseException;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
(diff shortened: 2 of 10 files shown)


andrewazores:
I'm thinking it will probably make sense to include some more details about the templates than just their names and the associated event map. Maybe the call to get their names should really return some collection of objects including the name but also description, etc.?

jiekang:
That seems reasonable to me though I'd rename name to template, so get templates and then get events by template name.

andrewazores:
`getNames` is now `getTemplates` and returns a struct of <name, description, provider> (although the "name" field really uses the "label" XML attribute, but that didn't seem a good fit for the API).

andrewazores:
Okay, after making those last changes and then the corresponding changes in `container-jfr` to test them out, I'm pretty happy with the current state. WIP label removed.

.gitignore
```diff
@@ -1,16 +1,6 @@
-.gradle
```
jiekang: I guess this is related to swapping from gradle to maven? Definitely pedantic but I'd prefer a separate commit for this. Can be the same PR.

andrewazores: Yea, that's correct. I'll split it out into a separate commit.

src/main/java/com/redhat/rhjmc/containerjfr/core/templates/RemoteTemplateService.java
```diff
@@ -0,0 +1,98 @@
...
+                                    m ->
+                                            m.getRoot().getAttributeInstances().stream()
+                                                    .anyMatch(
+                                                            attr ->
+                                                                    attr.getAttribute()
+                                                                                    .getName()
+                                                                                    .equals("label")
+                                                                            && attr.getValue()
+                                                                                    .equals(
+                                                                                            templateName)))
```
jiekang: Not a review comment, just curious. Is spotbugs applied here? There's a lot of indentation here heh...

andrewazores: Yea, that's autoformatted by spotless. Streaming collections and using lambdas like this tends to get a bit ugly because of those formatting rules.

jiekang: Cool, no worries

.gitignore
jiekang: Thank you!

PR #35, opened 2020-04-23
Add UPL license, managed by license-maven-plugin

ebaron:
This PR adds a LICENSE file to the root of the project, along with license headers to all Java source files. UPL was chosen to match the JMC project.


andrewazores:
Same comment as in rh-jmc-team/container-jfr#156 : can license-maven-plugin be applied here?

ebaron:
I've updated this to use license-maven-plugin, similar to https://github.com/rh-jmc-team/container-jfr/pull/156

PR #40, opened 2020-07-02
Local, Mutable Templates

andrewazores:
See also: rh-jmc-team/container-jfr#199

This PR extends the Templates API to allow the raw template XMLs to be retrieved, and to allow custom template XMLs to be added, parsed, validated, and saved to the local filesystem. These custom templates can be retrieved on their own, or in a "merged" view alongside the built-in templates for a specific remote target.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/core/templates/LocalStorageTemplateService.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/core/templates/LocalStorageTemplateService.java
@@ -49,6 +49,7 @@ import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import org.jsoup.Jsoup;
@@ -95,7 +96,7 @@ public class LocalStorageTemplateService extends AbstractTemplateService
     }
 
     @Override
-    public Document getXml(String templateName) throws FlightRecorderException {
+    public Optional<Document> getXml(String templateName) throws FlightRecorderException {
         for (Path path : getLocalTemplates()) {
             try (InputStream stream = fs.newInputStream(path)) {
                 Document doc =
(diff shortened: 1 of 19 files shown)


src/main/java/org/openjdk/jmc/flightrecorder/controlpanel/ui/ControlPanel.java
```diff
@@ -0,0 +1,58 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ * #L%
+ */
+package org.openjdk.jmc.flightrecorder.controlpanel.ui;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class ControlPanel {
```
jiekang, line 47: I guess this is needed to use the XMLModel stuff? :|

jiekang: Oh NVM, saw FIXME in the other file.

andrewazores: Yea, there's a static UIPlugin Logger instance reference somewhere in the JMC XMLModel code path - I don't remember which exact class off the top of my head - so this is another classloading hack "shim" to avoid a NoClassDefFound...

src/main/java/com/redhat/rhjmc/containerjfr/core/templates/LocalStorageTemplateService.java
```diff
@@ -0,0 +1,258 @@
...
+                        "Document contains multiple \"configuration\" elements");
+            }
+            Element configuration = els.first();
+            if (!configuration.hasAttr("label")) {
+                throw new InvalidXmlException(
+                        "Configuration element did not have \"label\" attribute");
+            }
+            // side effect of validation, throws ParseException or IllegalArgumentException if
+            // invalid
+            EventConfiguration.createModel(doc.toString());
```
jiekang, line 125: The setup parses through the xml twice, once with jsoup and once with the Java API, right? Could the XMLModel generated here be reused to accomplish what's needed for addTemplate? Are the checks made before this really needed? EventConfiguration is able to take an InputStream as well.

PR #45, opened 2020-07-29
Credentials implements equals/hashCode

andrewazores:
This allows Credentials and types which contain them (specifically, ContainerJFR's ConnectionDescriptor) to be used in Collection types like HashMaps. This is useful because a ConnectionDescriptor is a fairly natural caching key.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 <modelVersion>4.0.0</modelVersion>
 <groupId>com.redhat.rhjmc</groupId>
 <artifactId>containerjfr-core</artifactId>
-<version>0.14.0</version>
+<version>0.14.1</version>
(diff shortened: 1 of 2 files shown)


jiekang:
@andrewazores 
Is there anything particularly special about the lang3 builders that we can't incorporate into our own codebase? I'd prefer to rely on our own equals/hashcode implementations over a third-party library. Granted the dependency version we choose is "set in stone" and unlikely to change beneath us without our knowledge, it feels unnecessary.

andrewazores:
Not really, but we're already using commons-lang for a few other small purposes elsewhere in -core as well as a fair amount in container-jfr proper (including those same builders too), so it seemed like a nice and easy way to save a few lines for no additional dependency cost.

PR #46, opened 2020-08-26
Add Functions from JMC's ConnectionToolkit

### LINKED ISSUE #44: Add abstraction around JMC ConnectionToolkit
JMC has a useful utility class `ConnectionToolkit` with many static methods that would also be useful in ContainerJFR, ex. `createServiceURL`, `getDefaultPort`, `getHostName`, and perhaps the various `getXBean`s. A simple wrapper abstraction around this class should be created and added to -core so that it can be exposed to ContainerJFR while allowing -core to possibly reimplement the features independently of JMC in the future.

vic-ma:
Fixes #44.

These are the methods I didn't add. Would any of them be helpful to have?
```
public static ObjectName createObjectName(String name)
public static MemoryMXBean getMemoryBean(MBeanServerConnection server)
public static RuntimeMXBean getRuntimeBean(MBeanServerConnection server)
public static ThreadMXBean getThreadBean(MBeanServerConnection server)
public static OperatingSystemMXBean getOperatingSystemBean(MBeanServerConnection server)
public static Object invokeOperation(MBeanServerConnection server, ObjectName on, String operation, Object ... parameters)
public static boolean isJavaVersionAboveOrEqual(IConnectionHandle connectionHandle, JavaVersion minVersion)

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/core/net/JFRConnectionToolkit.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/core/net/JFRConnectionToolkit.java
@@ -90,7 +90,7 @@ public class JFRConnectionToolkit {
         return ConnectionToolkit.getHostName(url);
     }
 
-    public static int getPort(JMXServiceURL url) {
+    public int getPort(JMXServiceURL url) {
         return ConnectionToolkit.getPort(url);
     }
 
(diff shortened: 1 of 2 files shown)


andrewazores:
The various `getXBean` ones might be useful for something - in particular, the Memory/OperatingSystem ones could be helpful in https://github.com/rh-jmc-team/container-jfr/issues/8 . Those don't really need tests if they just directly call through to the JMC implementation.

vic-ma:
Are tests for the methods I currently have needed, or should I remove these too?

andrewazores:
The ones you wrote look good - I'm sure JMC has those methods also covered by unit tests, but if we were to reimplement this ourselves then these tests would help ensure correctness of the new implementation. The MXBean ones are less obvious how to test meaningfully.

andrewazores (approved):

PR #48, opened 2020-10-06
Avoid serializing credentials

andrewazores:
Mark credentials fields as transient to ensure serialization tools do not serialize the contents of Credentials objects.

https://sites.google.com/site/gson/gson-user-guide#TOC-Finer-Points-with-Objects


vic-ma:
Just needs a spotless format:
```
[ERROR]         -····private·transient·final·String·username;
[ERROR]         -····private·transient·final·String·password;
[ERROR]         +····private·final·transient·String·username;
[ERROR]         +····private·final·transient·String·password;
```

Everything seems to be working otherwise.

vic-ma (approved):

PR #49, opened 2020-10-23
Change Logger Implementation to Use SLF4J

### LINKED ISSUE #37: Logger should use SLF4J

vic-ma:
Fixes #37

Corresponding CJFR PR: rh-jmc-team/container-jfr#308

Regarding the problem of our dependencies using SLF4J and printing strange logs, the fix I did was to disable logging for JMC and Vertx classes.

The reason I disabled Vertx logs as well is that I found that I was getting an NPE log from Vertx when using the web client and starting a recording. It's from the end-handler logger call (that prints the path of the request, status of the response, etc.) of a report-get request, and the thing that is null is `req.remoteAddress()`.

The null pointer is present in upstream too, but the logger just doesn't print the exception for some reason (it doesn't print the log at all). I checked by changing the logger call in upstream to a `printf()` call that sets the field to "NULL" if it's null (so it's more readable, since there isn't a big exception log). It's the fifth line from the bottom: 
```
(Right after pressing "Create" in the web-client recording creation form)

[TRACE] Locking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
[TRACE] Unlocking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
(127.0.0.1:45302): POST /api/v1/targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcontainer-jfr%3A9091%2Fjmxrmi/recordings 201 935ms
(127.0.0.1:45302): GET /api/v1/targets 200 2ms
[TRACE] Locking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
[TRACE] Unlocking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
(127.0.0.1:45302): GET /api/v1/targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcontainer-jfr%3A9091%2Fjmxrmi/recordings 200 116ms
(127.0.0.1:45306): GET /api/v1/recordings 200 0ms
[TRACE] Locking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
[TRACE] Unlocking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
(127.0.0.1:45302): GET /api/v1/targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcontainer-jfr%3A9091%2Fjmxrmi/recordings 200 86ms
(127.0.0.1:45306): GET /api/v1/recordings 200 0ms
(NULL): GET /api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcontainer-jfr:9091%2Fjmxrmi/reports/foo 200 26ms
[TRACE] Locking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
[TRACE] Active report cache miss for foo
[TRACE] Unlocking connection service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi
(127.0.0.1:45344): GET /api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcontainer-jfr:9091%2Fjmxrmi/reports/foo 200 1410ms
```

I'm still a bit confused about how `LogManager.reset()` worked, since it says that it sets every logger's level to `null` and sets the root to `INFO`; but wouldn't that just make everything `INFO`, rather than disabling them?

I also don't get why the NPE doesn't get printed in upstream, since the binding we used is slf4j-simple, and I don't think there's any configuration that should disable that message.


pom.xml
```diff
@@ -5,7 +5,7 @@
 <modelVersion>4.0.0</modelVersion>
 <groupId>com.redhat.rhjmc</groupId>
 <artifactId>containerjfr-core</artifactId>
-<version>0.15.0</version>
+<version>0.15.1</version>
```
andrewazores, line 8: Let's make this `0.16.0`

src/main/java/com/redhat/rhjmc/containerjfr/core/ContainerJfrCore.java
```diff
@@ -52,11 +56,22 @@
...
-        LogManager.getLogManager().reset();
+
+        String jmcLoggerLevelProperty = "org.openjdk.jmc.level=" + Level.OFF.getName();
+        String vertxLoggerLevelProperty = "io.vertx.level=" + Level.OFF.getName();
+
+        String config = jmcLoggerLevelProperty + "\n" + vertxLoggerLevelProperty;
+
+        LogManager.getLogManager()
+                .updateConfiguration(
+                        new ByteArrayInputStream(config.getBytes(StandardCharsets.US_ASCII)),
```
andrewazores, line 72: This `ByteArrayInputStream` should be closed just as a matter of code cleanliness, IMO, even though closing this type of stream does nothing. Please wrap it in a try-with-resources.

src/main/java/com/redhat/rhjmc/containerjfr/core/ContainerJfrCore.java
```diff
@@ -52,11 +56,22 @@
...
+            throws CoreException, SecurityException, NullPointerException, IOException {
         System.setProperty(
                 "org.openjdk.jmc.common.security.manager",
                 SecurityManager.class.getCanonicalName());
-        LogManager.getLogManager().reset();
+
+        String jmcLoggerLevelProperty = "org.openjdk.jmc.level=" + Level.OFF.getName();
+        String vertxLoggerLevelProperty = "io.vertx.level=" + Level.OFF.getName();
+
+        String config = jmcLoggerLevelProperty + "\n" + vertxLoggerLevelProperty;
```
andrewazores, line 68: Rather than hardcoding this config here, could the configuration instead be read from a file? The file could be specified by a system property or env var, and if unset, default to these hardcoded values. This way, if these log messages might ever be useful for debugging something, they can be enabled by the deploying end user by simply setting a property/env var and pointing it at a blank configuration file to override this configuration - or pointed at a file that sets the levels to something other than `OFF`.

vic-ma: I could change the mapper function to use the old value if it exists, instead of using the new value if it exists (so preference is given to the old value in the case where both exist). That way the user can override these defaults by just setting the same properties in a config file, and pass that in via `-Djava.util.logging.config.file`. (This is the way I had it before, but I went back and forth on it.) And okay, I'll move the defaults to an external file and probably link to it from the README so the user can see which properties they might want to override.

vic-ma: I realised that passing in the path to a config file via an env var that sets `-Djava.util.logging.config.file` won't work inside a container. So I think I'll change it so that the env var is the configuration itself, and then just call `updateConfiguration()` function on it.

andrewazores: Well, it still could work, but the user would need to mount the config file to the container. That's not really that big of a deal.

vic-ma: I think I've got both ways working now. Is there one you'd prefer?

andrewazores: I think I prefer the file route, simply because setting potentially multiline and long settings like that is a bit unconventional (although it should work) with environment variables.

vic-ma:
The new env var basically just sets `-Djava.util.logging.config.file`. The `updateConfiguration()` function now prefers the old config.

To test the custom config function, I just took Java's default `logging.properties` and then added at the end:
```
org.openjdk.jmc.level=ALL
io.vertx.level=ALL
```
And creating a recording would generate a lot of error logs.

andrewazores:
Where/what is the new env var?

vic-ma:
I thought I fixed this but the `SecurityException` and `NullPointerException` in the throws statement should be unncessary, so I've removed them.

andrewazores (approved):
Looks good. For a follow-up I think we should think about removing the singleton Logger instance and implement something like `Logger.getLogger(Class<?> klazz)` so that things logged by ContainerJFR can be more easily filtered by which class emitted the log.

PR #54, opened 2020-12-17
feat(filesystem): add abstractions for creating (temp) files/dirs

andrewazores:
Fixes https://github.com/cryostatio/cryostat-core/issues/113

### DIFF
--- a/src/main/java/io/cryostat/core/sys/FileSystem.java
+++ b/src/main/java/io/cryostat/core/sys/FileSystem.java
@@ -116,8 +116,8 @@ public class FileSystem {
         return Files.createFile(path, attrs);
     }
 
-    public Path createTempFile(String prefix, String suffix, FileAttribute<?>... attrs) throws
-        IOException {
+    public Path createTempFile(String prefix, String suffix, FileAttribute<?>... attrs)
+            throws IOException {
         return Files.createTempFile(prefix, suffix, attrs);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/cryostat/core/sys/FileSystem.java
```diff
@@ -110,4 +111,35 @@ public Path pathOf(String first, String... more) {
...
+
+    public Path createDirectories(Path path, FileAttribute<?>... attrs) throws IOException {
+        return Files.createDirectories(path, attrs);
+    }
+
+    public Path createTempDirectory(String prefix, FileAttribute<?>... attrs) throws IOException {
+        return Files.createTempDirectory(prefix, attrs);
+    }
+
+    public Path createTempDirectiry(Path path, String prefix, FileAttribute<?>... attrs)
```
hareetd, line 141: "Directory" is misspelled here.

hareetd:
Do we wrap these methods from `java.nio.file.Files` inside our own methods in order to decouple the implementation details from any calling classes (i.e. we could change the library we use for the actual file operations without having to change all the different places these methods get called throughout the Cryostat backend)?

andrewazores:
No, we would probably never in effect replace the underlying implementation. This will always just be a wrapper `java.nio.Files`, I think. The reason for doing it this way was primarily just so that we can inject a `FileSystem` object and use its instance methods, rather than calling static methods on the `Files` class and others, and this makes testing easier since we can just inject a mock FileSystem and have that do whatever we need. There are ways to mock or stub static method calls using PowerMockito. or newer versions of Mockito 3.4.0+, but this abstraction/decoupling pattern in Cryostat is from before Mockito 3.4.0, and we would have needed to bring in PowerMockito or another dependency just to avoid this simple injection pattern.

andrewazores:
There are also some other complexities and easy mistakes to make/traps to fall into when mocking statics, even now that it's possible with normal Mockito's latest versions, so perhaps it's still best to avoid doing that when possible.

https://www.baeldung.com/mockito-mock-static-methods is a nice article showing what mocking statics looks like - it's messier than mocking normal instances.

hareetd:
Ah okay, yeah that article definitely shows the downsides.

hareetd (approved):

PR #62, opened 2021-03-30
Remove CHANGED discovery event

andrewazores:
Remove CHANGED discovery event which mirrored JMC API but is not generally applicable. Instead, a pair of LOST/FOUND is now emitted to inform clients of any changes to targets.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/core/net/discovery/JvmDiscoveryClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/core/net/discovery/JvmDiscoveryClient.java
@@ -48,7 +48,8 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


Josh-Matsuoka:
Hey Andrew,

Changes look good to me. The Lost/Found event behavior is as expected when testing locally with podman, I'm unable to reproduce a Changed event though.

Josh-Matsuoka (approved):

andrewazores:
@Alexjsenn do you remember the details of the `CHANGED` event and how it's (theoretically) triggered? I remember you looked into it before but I don't recall if you had actually found a real scenario where it is fired.

Alexjsenn:
> @Alexjsenn do you remember the details of the `CHANGED` event and how it's (theoretically) triggered? I remember you looked into it before but I don't recall if you had actually found a real scenario where it is fired.

I don't remeber the exact details, but I recall there was no concrete case were we thought we would need to support it as such. If you need I can look into it again, I can't recall exactly how it was triggered.

PR #80, opened 2021-04-16
Add AdoptOpenJDK Maven repo

andrewazores:
Removes need to manually build JMC locally into .m2 repository

Now that the non-core JMC pieces we use in -core have been copied into our tree, even if temporarily, we can avoid the requirement of manually building JMC at all by adding the AdoptOpenJDK repo for JMC Core to our -core's pom. Eventually the JMC Core dependencies should end up on Maven Central as well, so this repository definition can be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,8 @@ JFR with JDK Mission Control API
 Build:
 - Maven
 - JDK11+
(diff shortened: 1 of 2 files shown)


ebaron (approved):
Works as expected. Very nice improvement!

PR #85, opened 2021-04-20
Rename to cryostat-core

andrewazores:


src/main/java/io/cryostat/core/net/Credentials.java
```diff
@@ -39,26 +39,26 @@
...
 
-    String getUsername() {
+    public String getUsername() {
         return username;
     }
 
-    String getPassword() {
+    public String getPassword() {
         return password;
     }
```
ebaron, line 63: Is this visibility change intentional?

ebaron (commented):
Looks good, just one question about a non-renaming change.

src/main/java/io/cryostat/core/net/Credentials.java
andrewazores: No, that looks like a mistake - not sure how that snuck in. Thanks for catching it, very sharp eye!

andrewazores: (that change is used for something else in a testing branch related to autorules, but I don't know how it ended up in this patch)

ebaron (approved):

PR #91, opened 2021-04-22
Update license plugin and copyright headers

andrewazores:


ebaron (commented):
This should probably exclude the embedded JMC files.

ebaron (approved):
Looks good!

PR #108, opened 2021-12-31
feat(reportgenerator): implement cancellable report generator

### LINKED ISSUE #109: Report Generation is long-running but cannot be cancelled

andrewazores:
Fixes #109


src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
```diff
@@ -0,0 +1,350 @@
...
+                    });
+            return document.outerHtml();
+        } catch (Exception e) {
+            logger.warn(e);
+            return report;
+        }
+    }
+
+    /**
+     * Everything below this point is forked from JMC's JfrHtmlRulesReport and RulesToolkit classes,
```
jan-law, line 177: Out of curiosity, how would you know whether or not to update cryostat-core if jmc makes changes to these method implementations?

andrewazores: It's totally manual and best-effort at this point. If this continues to compile and our integration tests pass then we can feel pretty safe that nothing was fundamentally broken, but if there are any behavioural changes under the hood in JMC, that's just up to me/maintainers to review the changes and adapt our -core as needed. So if the JMC classes we're relying on become interruptible on their own to begin with, or they refactor their other existing cancellable/interruptible implementation to not be tightly coupled to Eclipse, then we could just drop my implementation here for example.

ebaron: Do you think any of these changes make sense to integrate into upstream JMC in the future?

andrewazores: Not sure. AFAICT this entrypoint into the rules processing system isn't really used by JMC anymore - they appear to just be using the Eclipse-coupled entrypoint instead. I'm not sure of the history between those two, or if I'm missing a place where the entrypoint we use is actually still used by JMC somewhere.

If it is valuable to JMC at all then I would certainly be happy to upstream the change.

src/test/java/io/cryostat/core/reports/InterruptibleReportGeneratorTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Mock InputStream recording;
+    Set<ReportTransformer> transformers;
+    InterruptibleReportGenerator generator;
+
+    @BeforeEach()
+    void setup() throws Exception {
+        transformers = new HashSet<>();
+        generator =
+                new InterruptibleReportGenerator(
+                        logger, transformers, Executors.newWorkStealingPool(1));
```
jan-law, line 76: Why did you choose to use `Executors.newWorkStealingPool()` here, while the `interruptible-reports` branch uses `ForkJoinPool.CommonPool()` in https://github.com/cryostatio/cryostat-reports/commit/6474ffeb646f57a3896e8a35d396200aec054495?

andrewazores: In actual use in that -reports commit:

> A static commonPool() is available and appropriate for most applications. The common pool is used by any ForkJoinTask that is not explicitly submitted to a specified pool. Using the common pool normally reduces resource usage (its threads are slowly reclaimed during periods of non-use, and reinstated upon subsequent use). 

(https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ForkJoinPool.html)

For the unit test I used `newWorkStealingPool(1)` so that the test has its own single and independent thread that behaves similarly to one of the threads that might be found in the `ForkJoinPool`. The test could have used the `commonPool` as well, but I thought it made sense for the unit test to be as simple and predictable as possible, so all tests are run on a single thread rather than possibly being spread across multiple threads in the common pool.

jan-law (commented):
Looks good to me! @ebaron @hareetd what do you think?

ebaron (approved):
Changes look fine to me

src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
```diff
@@ -0,0 +1,350 @@
...
+    public Future<String> generateReportInterruptibly(InputStream recording) {
+        return qThread.submit(
+                () -> {
+                    // this is generally a re-implementation of JMC JfrHtmlRulesReport#createReport,
+                    // but calling our cancellable evalute() method rather than the
+                    // RulesToolkit.evaluateParallel as explained further down.
+                    List<Future<Result>> resultFutures = new ArrayList<>();
+                    try {
+                        resultFutures.addAll(
+                                evaluate(
```
hareetd, line 109: I realize this has been already merged but I have a couple of questions.

So when `evaluate` is called with the list of `rules` in the registry and the actual `events` content of the recording in question, the recording is evaluated against each `rule` and packaged into a `RunnableFuture<Result>`, which gets `run` and the `Result` from calling `get` packaged into a `Callable<Result>`. A list of these is returned, streamed, and mapped to `executor::submit`, which spits out a cancellable `Future<Result>`.

What I don't understand is didn't we already wait for the rule evaluation to complete inside `evaluate` itself, specifically when `return result.get()` was called on line 236?

Also, am I correct in saying that the way report generation works is by evaluating the recording against a bunch of standard rules for proper JVM behaviour, the results for which (i.e. any errors) are then mapped to certain result groups, which allows for a clear, organized report once everything has been processed to HTML, correct?

hareetd: Actually, I thought about it more and realized that when the lambda function inside `evaluate` is packaged into a `Callable<Result>`, it happens asynchronously so I think the actual rule evaluation isn't triggered until `map(executor::submit)` is called on line 113.

andrewazores: On line 236 we are doing a blocking `.get()` call from within a task submitted to the thread pool, so these are being processed asynchronously on a different thread than the main one that is first handling the request. If there is only one CPU core available then this basically just degenerates to doing it sequentially, but it's still a separate thread, and so we can interrupt it and have the CPU context-switch back to the main thread and stop processing any rules.

> Also, am I correct in saying that the way report generation works is by evaluating the recording against a bunch of standard rules for proper JVM behaviour, the results for which (i.e. any errors) are then mapped to certain result groups, which allows for a clear, organized report once everything has been processed to HTML, correct?

Yep, pretty much. If you look at an HTML rules report, each of those coloured score bar categories that you see is a rule. If you have enough CPU cores then all of those can be evaluated in parallel. Once they are evaluated then the result score is used to generate the HTML for the coloured bar, the associated text is placed next to it, related rules are grouped together, and all of those are finally combined into the full page HTML report document.

andrewazores: Or rather - on line 236 we are converting the `RunnableFuture` representing the `Rule` into a simpler `Callable`, which both performs the execution of the rule and returns its result. Line 113, `executor::submit`, is what actually turns those `Callables` into asynchronous tasks processed by the thread pool.

hareetd: Awesome, makes sense. Thanks!

PR #118, opened 2022-01-24
feat(reports): Add JFR event for report rules

### LINKED ISSUE #117: Record a JFR event when evaluating report rules
> // TODO it would be very nice to record a JFR event here to pick up in
// Cryostat's own profiling template, to compare the different rule
// implementations and the time they each take to process with various
// source recording settings

https://github.com/cryostatio/cryostat-core/blob/main/src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java#L231

Related to https://github.com/cryostatio/cryostat/issues/461

I can work on this.

_Originally posted by @jan-law in https://github.com/cryostatio/cryostat/issues/461#issuecomment-1018895266_

jan-law:
Fixes #117

Question: How do I test this change with my other branch in `~/cryostat`? I tried the below commands on my `record-rule-event` branches, started smoketest, started a new recording on `service:jmx:rmi:///jndi/rmi://cryostat:10000/jmxrmi` with the updated Cryostat template, generated a report, then downloaded the recording. I couldn't find any instances of my custom event in the recording file with `jfr print <recording_name>.jfr | vim -`. What am I missing?

```
~/cryostat-core $ mvn install
~/cryostat-core $ cd ~/cryostat
~/cryostat $ mvn clean package && CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true sh smoketest.sh
```

### DIFF
--- a/src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
+++ b/src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
@@ -69,6 +69,11 @@ import org.openjdk.jmc.flightrecorder.rules.report.html.internal.RulesHtmlToolki
 
 import io.cryostat.core.log.Logger;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import jdk.jfr.Category;
+import jdk.jfr.Event;
+import jdk.jfr.Label;
+import jdk.jfr.Name;
 import org.jsoup.Jsoup;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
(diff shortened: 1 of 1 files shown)


andrewazores:
https://github.com/cryostatio/cryostat/blob/7361dfee2ed32920171ca6219406c94086d5f436/src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java#L286

The SubprocessReportGenerator doesn't use `InterruptibleReportGenerator`, just `-reports` does.

I think in your main Cryostat PR you can just update the `SubprocessReportGenerator` to use `InterruptibleReportGenerator` so that this event gets emitted. There should be little to no other functional change to swapping that implementation.

andrewazores (approved):

PR #120, opened 2022-01-25
fix(report): register rule eval JFR event on init

### LINKED ISSUE #119: Rule report evaluation event should be registered

andrewazores:
Fixes #119

### DIFF
--- a/src/main/java/io/cryostat/core/CryostatCore.java
+++ b/src/main/java/io/cryostat/core/CryostatCore.java
@@ -42,6 +42,9 @@ import java.io.InputStream;
 import java.util.logging.LogManager;
 
 import io.cryostat.core.jmc.SecurityManager;
+import io.cryostat.core.reports.InterruptibleReportGenerator.ReportRuleEvalEvent;
+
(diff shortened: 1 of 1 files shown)


jan-law:
Could you explain why this event needs to be manually registered compared to the events in cryostat -base?

andrewazores:
Events are supposed to be automatically registered by default when their class is loaded - we can see this happens with the other events you and Hareet have added in -base, because there is no specific configuration for those, and yet they can be seen in the "Event Types" view for the Cryostat JVM. This indicates that the JVM's JFR framework does know that those events do exist - maybe because by the time a Cryostat instance has started up and gotten to the point that we can make such an API request, all of those events will have been emitted already, and so the classloader would have loaded them and whatever internal JFR mechanism has caused them to auto-register.

On the other hand, this report generation event should not be emitted in Cryostat's application startup phase, and so the class will not be loaded and the event will not be registered with JFR, unless we either trigger it to be emitted and registered or we explicitly register it in advance. Once it is registered then it appears in the "Event Types" view like the rest.

jan-law (approved):

PR #121, opened 2022-01-27
feat(reports): Add entire report generation JFR event

jan-law:
Related #117 

This event records the entire report generation duration for both the reports sidecar replica and the subprocess report generator.


jan-law:
> WDYT about wrapping the provided stream in something like https://commons.apache.org/proper/commons-io/javadocs/api-2.4/index.html?org/apache/commons/io/input/CountingInputStream.html , so that the event can also record the size of the recording?

That's a good idea that would help us compare the relationship between recording sizes and the time it takes to process them. I'll add that in.

> You could also iterate over the Results and check their score. Anything with a score of -1 was processed as NOT_APPLICABLE (https://github.com/openjdk/jmc7/blob/7db57f0b7f72020db4afd17f84d71c5f1f1903d9/core/org.openjdk.jmc.flightrecorder.rules/src/main/java/org/openjdk/jmc/flightrecorder/rules/Result.java#L48), so we can use that as a proxy for the "complexity" of the recording.

If I understood correctly, reports with fewer applicable `Results` are easier to process, so we could count the number of valid scores and assign a "complexity" value to the event. And a report is considered more complex if it has a greater number of applicable results?

andrewazores:
I think just recording the total number of `Result`s (this can change between JMC versions as more rules are added to the framework), and the number that had a `-1` score indicating that no processing was really done for that rule, nothing beyond checking that any of the events it requires were present in the source recording. So the event will look like

```
Event {
  String recordingName
  int rulesEvaluated
  int rulesApplicable
  long recordingSize
}
```

andrewazores:
The complexity isn't a direct metric to record I don't think, since what I conceptualize it as right now is basically just that number of rules processed that actually "did something" - but the number of rules that were even checked to begin with, and the work that each rule does, is likely to change between JMC versions. So it's best not to try to boil down the raw values into a single "complexity score" here, because then older recordings vs newer recordings made with different JMC versions behind the -core implementation will change the complexity score in ways that make them hard to compare.

jan-law:
If I want to begin a `ReportGenerationEvent` in the `SubprocessReportGenerator` `exec()` call so that Cryostat can access the JFR event, how can I update `rulesEvaluated` or `rulesApplicable` to this event while the `InterruptibleReportGenerator` is computing the `Results` in [generateReportInterruptibly()](https://github.com/cryostatio/cryostat-core/blob/main/src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java#L120)? Isn't the event scoped to the function it was declared in?

andrewazores:
Right, you wouldn't be able to do it directly - the event needs to be created by, reside in, and be committed by the main Cryostat JVM process and not the subprocess, or else we can't easily record the event. So that means that the event fields can't be updated by the subprocess.

One pretty simple way around this would be to have the subprocess check those statistics and write them out to a temporary file before it exits, at some well-known/constant filesystem location. Cryostat can check if that file exists after the subprocess exits, and if it does it can read the file and use the contents to update the event fields. Be sure to end the event before doing this extra file I/O so that this additional overhead doesn't count against the execution time for the report generation.

Very rough illustrative example:

```
cryostatJvm {
  exec() {
    ReportGenerationEvent evt = new ReportGenerationEvent();
    try {
      evt.start();
      Future<String> reportFileContent = forkSubprocess(args).get();
      evt.end();
      Map<String, String> stats = gson.fromJson(readFile(STATS_PATH));
      evt.rulesEvaluated = stats.get("rulesEvaluated");
    } finally {
      if (evt.shouldCommit()) {
        evt.commit();
      }
    }
}

subprocess {
  main(String[] args) {
      CountingInputStream recording = new CountingInputStream(new BufferedInputStream(args[0]));
      String report = generator.generateReport(recording);
      long recordingSize = recording.getCount();
      writeFile(new Gson().toJson(Map.of("recordingSize", recordingSize)), STATS_PATH);
  }
}
```

andrewazores:
Hmm, good point. It's definitely a bit of a pain working with events at the library level like this.

Maybe we can update the `InterruptibleReportGenerator` so that its `generateReportInterrutibly` doesn't just return a `Future<String>`, but something like a `Future<ReportResult>`, where that looks like:

```
ReportResult {
  String html;
  long sourceRecordingLength;
  int rulesEvaluated;
  int rulesApplicable;
}
```

At the library level here in -core it can both return those stats as part of the method return, as well as record them in a JFR event.

Then, when the subprocess generator calls the function, it takes the returned stats values and passes those back to its parent process. The simplest way to do that is probably just to write them out (as JSON perhaps) to the filesystem at some well-known/constant location. After the subprocess exits then the main Cryostat parent process can check that filesystem location, and if it finds a file there it can read it to extract the stats data, and then use those to update the ReportGenerationEvent that it's going to record, and delete the stats file when it's finished.

andrewazores:
`InterruptibleReportGenerator`
```
Future<ReportResult> generateReportInterruptibly(InputStream stream) {
  CountingInputStream countingStream = new CountingInputStream(stream);
  // do everything else as usual
  return new ReportResult(report, countingStream.getCount());
}
```

`SubprocessReportGenerator`
```
mainCryostatJvm {
  exec() {
    ReportGenerationEvent evt = new ReportGenerationEvent(recordingName);
    try {
      evt.start();
      String report = forkSubprocess(stream).get();
      evt.end();
      if (fs.fileExists(STATS_PATH)) {
        Map<String, String> stats = gson.fromJson(new FileInputStream(STATS_PATH));
        evt.recordingLength = stats.get("recordingLength");
        fs.delete(STATS_PATH);
      }
    } finally {
      if (evt.shouldCommit()) evt.commit();
  }

subprocess {
  main(args) {
    ReportResult reportResult = generateReport(new FileInputStream(args[0]));
    writeFile(reportResult.html, reportOutputPath);
    writeFile(gson.toJson(Map.of("recordingLength", reportResult.recordingLength), STATS_PATH);
  }
}
```

andrewazores:
This way the subprocess generation can collect the stats and record them to JFR in a fairly straightforward way, and the remote reports generation sidecar can just use the JFR event directly emitted in the -core library code which contains the same stats data.

andrewazores:
This PR and the two associated ones are looking good, I think. Please rebase them all. I'm going to keep thinking about how we might better do version updates for `-core` that don't require a new minor version release every time we want to add something.

andrewazores:
Rebase please. I think for now I'll just stick to keeping the minor version number release scheme, so your other two dependent PRs will need an update to increment the version number later on.

andrewazores:
@jan-law 2.7.0 should be usable for your dependent PRs now

Issue #136, opened 2022-05-03
[Task] Report generator should allow callers to specify which rules to evaluate


andrewazores:
Core implementation of https://github.com/cryostatio/cryostat/issues/918

PR #138, opened 2022-05-30
feat(reports): report generator allows callers to specify which rules to evaluate

turkeychikken:
Related to #136

Not sure if this is all we need to allow callers to specify rules because all it is, is a method overload with an extra parameter added. Feedback is appreciated.

I was thinking maybe if the Collection of rules is empty or null that there would be some custom exception that would need to be implemented? But I am also confused as to how to test cryostat-core implementations in general.

### DIFF
--- a/src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
+++ b/src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
@@ -45,6 +45,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -52,6 +53,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.RunnableFuture;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.openjdk.jmc.common.io.IOToolkit;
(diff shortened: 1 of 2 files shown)


andrewazores:
> Not sure if this is all we need to allow callers to specify rules because all it is, is a method overload with an extra parameter added. Feedback is appreciated.

Yep, pretty much. Though requiring the caller to pass in the whole `IRule` seems potentially burdensome. Could that be a `Collection<String>`, where the String is just the rule name for example? It might also be useful to let callers specify a collection of rule topics to evaluate.

> I was thinking maybe if the Collection of rules is empty or null that there would be some custom exception that would need to be implemented?

If the collection is empty I think that just results in no evaluations and a report where every score is N/A. If it's null then that's just a `NullPointerException`.

> But I am also confused as to how to test cryostat-core implementations in general.

Unit tests, primarily :-)

If you want to test it out "live", update your `pom.xml` locally to bump the version number, and then do a `mvn install` to build your changes with that new version number into your local Maven repository. Then you can go into ex. `cryostat-reports` and edit the `pom.xml` there to require your updated `-core` version, and build some new functionality around your `-core` changes.

turkeychikken:
> It might also be useful to let callers specify a collection of rule topics to evaluate.
What do you mean by this, because I think that's what I was doing before. Are you saying that there should be Collection<String> as well as there could be Collection<IRule> ?

andrewazores:
Yea, I think `Collection<IRule>` is inconvenient for callers to use. The caller doesn't have the `IRule` objects in hand, they just know the name of the rule or the topic it's associated with.

From the perspective of wanting to call into this method starting from an HTTP API request, how is the HTTP API request supposed to encode the whole `IRule` object to specify it should be evaluated? I think it's easier to just have the API request include a list of rule names than a list of `IRule`s, for example.

https://github.com/openjdk/jmc/blob/master/core/org.openjdk.jmc.flightrecorder.rules/src/main/java/org/openjdk/jmc/flightrecorder/rules/IRule.java

turkeychikken:
feedback? I think this is just what you showed me before but hopefully I didn't somehow mess it up somewhere, ignore the spotless checks and stuff for now

src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
```diff
@@ -104,17 +106,27 @@ public InterruptibleReportGenerator(
...
+        return generateReportInterruptibly(recording, rule -> true);
+    }
+
+    public Future<ReportResult> generateReportInterruptibly(
+            InputStream recording, Predicate<IRule> predicate) {
         return qThread.submit(
                 () -> {
-                    // this is generally a re-implementation of JMC JfrHtmlRulesReport#createReport,
-                    // but calling our cancellable evalute() method rather than the
-                    // RulesToolkit.evaluateParallel as explained further down.
```
andrewazores, line 111: I think this comment is worth keeping around

src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
```diff
@@ -104,17 +106,27 @@ public InterruptibleReportGenerator(
...
                     List<Future<Result>> resultFutures = new ArrayList<>();
                     try (CountingInputStream countingRecordingStream =
                             new CountingInputStream(recording)) {
+                        Objects.requireNonNull(predicate);
```
andrewazores, line 119: Probably better to perform the null-check before `qThread.submit` and before opening the recording stream. If the request is invalid, may as well fail ASAP.

src/main/java/io/cryostat/core/reports/InterruptibleReportGenerator.java
```diff
@@ -104,17 +106,28 @@ public InterruptibleReportGenerator(
...
+
+    public Future<ReportResult> generateReportInterruptibly(
+            InputStream recording, Predicate<IRule> predicate) {
         return qThread.submit(
                 () -> {
                     // this is generally a re-implementation of JMC JfrHtmlRulesReport#createReport,
                     // but calling our cancellable evalute() method rather than the
                     // RulesToolkit.evaluateParallel as explained further down.
+                    Objects.requireNonNull(predicate);
+                    Objects.requireNonNull(recording);
```
andrewazores, line 120: It would still be better to move these outside of the `qThread`, I think. The failure will occur a little sooner, and you'll probably get a cleaner stack trace out of it too.

turkeychikken: Oh, yeah I don't know why but I was thinking that this was going on a new thread so that the extra null checks wouldnt affect the main thread, but that doesnt make sense now that I see it.

andrewazores (commented):
Looks good. Could you add tests that verify the NPE behaviour when the recording and/or predicate are null?

andrewazores (approved):

PR #141, opened 2022-06-01
feat(jmx): compute unique jvm id

jan-law:
Related https://github.com/cryostatio/cryostat/issues/960
Related https://github.com/cryostatio/cryostat-web/issues/442

~~Adds a getter that returns the JFRConnection server descriptor's GUID.~~

Adds a getter that computes a hashcode based on jvm runtime properties.

### DIFF
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -186,7 +186,7 @@ public class JFRConnection implements AutoCloseable {
                 } else {
                     throw new IllegalArgumentException(
                             String.format(
-                                    "Unexpected attribute class of %s", attrObject.getClass()));
+                                    "Unexpected attribute: {%s} of type: {%s}", attr, attrObject.getClass()));
                 }
             }
             byte[] hash = DigestUtils.sha256(baos.toByteArray());
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -76,7 +76,6 @@ import io.cryostat.core.templates.TemplateService;
 import io.cryostat.core.tui.ClientWriter;
 
 import org.apache.commons.codec.digest.DigestUtils;
-import org.apache.commons.lang3.ArrayUtils;
 
 public class JFRConnection implements AutoCloseable {
 
@@ -186,10 +185,8 @@ public class JFRConnection implements AutoCloseable {
                         this.rjmxConnection.getAttributeValue(
                                 new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
                 if (attrObject.getClass().isArray()) {
-                    if (attrObject.getClass().getComponentType().isPrimitive()) {
-                        attrObject = castPrimitiveToObject(attrObject);
-                    }
-                    dos.writeUTF(Arrays.toString((Object[]) attrObject));
+                    String stringified = stringifyArray(attrObject);
+                    dos.writeUTF(stringified);
                 } else {
                     dos.writeUTF(attrObject.toString());
                 }
@@ -199,47 +196,46 @@ public class JFRConnection implements AutoCloseable {
         }
     }
 
-    private synchronized Object[] castPrimitiveToObject(Object primitiveObject) {
-        Object[] objArr;
-        String componentType = primitiveObject.getClass().getComponentType().toString();
+    private String stringifyArray(Object arrayObject) {
+        String stringified;
+        String componentType = arrayObject.getClass().getComponentType().toString();
         switch (componentType) {
             case "boolean":
-                objArr = ArrayUtils.toObject((boolean[]) primitiveObject);
+                stringified = Arrays.toString((boolean[]) arrayObject);
                 break;
 
             case "byte":
-                objArr = ArrayUtils.toObject((byte[]) primitiveObject);
+                stringified = Arrays.toString((byte[]) arrayObject);
                 break;
 
             case "char":
-                objArr = ArrayUtils.toObject((char[]) primitiveObject);
+                stringified = Arrays.toString((char[]) arrayObject);
                 break;
 
             case "short":
-                objArr = ArrayUtils.toObject((short[]) primitiveObject);
+                stringified = Arrays.toString((short[]) arrayObject);
                 break;
 
             case "int":
-                objArr = ArrayUtils.toObject((int[]) primitiveObject);
+                stringified = Arrays.toString((int[]) arrayObject);
                 break;
 
             case "long":
-                objArr = ArrayUtils.toObject((long[]) primitiveObject);
+                stringified = Arrays.toString((long[]) arrayObject);
                 break;
 
             case "float":
-                objArr = ArrayUtils.toObject((float[]) primitiveObject);
+                stringified = Arrays.toString((float[]) arrayObject);
                 break;
 
             case "double":
-                objArr = ArrayUtils.toObject((double[]) primitiveObject);
+                stringified = Arrays.toString((double[]) arrayObject);
                 break;
 
             default:
-                throw new IllegalStateException(
-                        String.format("Unexpected primitive array of type %s", componentType));
+                stringified = Arrays.toString((Object[]) arrayObject);
         }
-        return objArr;
+        return stringified;
     }
 
     public synchronized boolean isV1() {
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -202,40 +202,42 @@ public class JFRConnection implements AutoCloseable {
     private synchronized Object[] castPrimitiveToObject(Object primitiveObject) {
         Object[] objArr;
         String componentType = primitiveObject.getClass().getComponentType().toString();
-        switch(componentType) {
+        switch (componentType) {
             case "boolean":
-                objArr = ArrayUtils.toObject((boolean []) primitiveObject);
+                objArr = ArrayUtils.toObject((boolean[]) primitiveObject);
                 break;
-                
+
             case "byte":
-                objArr = ArrayUtils.toObject((byte []) primitiveObject);
+                objArr = ArrayUtils.toObject((byte[]) primitiveObject);
                 break;
 
             case "char":
-                objArr = ArrayUtils.toObject((char []) primitiveObject);
+                objArr = ArrayUtils.toObject((char[]) primitiveObject);
                 break;
 
             case "short":
-                objArr = ArrayUtils.toObject((short []) primitiveObject);
+                objArr = ArrayUtils.toObject((short[]) primitiveObject);
                 break;
-            
+
             case "int":
-                objArr = ArrayUtils.toObject((int []) primitiveObject);
+                objArr = ArrayUtils.toObject((int[]) primitiveObject);
+                break;
 
             case "long":
-                objArr = ArrayUtils.toObject((long []) primitiveObject);
+                objArr = ArrayUtils.toObject((long[]) primitiveObject);
                 break;
-            
+
             case "float":
-                objArr = ArrayUtils.toObject((float []) primitiveObject);
+                objArr = ArrayUtils.toObject((float[]) primitiveObject);
                 break;
-            
+
             case "double":
-                objArr = ArrayUtils.toObject((double []) primitiveObject);
+                objArr = ArrayUtils.toObject((double[]) primitiveObject);
                 break;
-                
+
             default:
-                throw new IllegalStateException(String.format("Unexpected primitive array of type %s", componentType));
+                throw new IllegalStateException(
+                        String.format("Unexpected primitive array of type %s", componentType));
         }
         return objArr;
     }
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -148,6 +148,10 @@ public class JFRConnection implements AutoCloseable {
         }
     }
 
+    public synchronized String getGUID() {
+        return this.rjmxConnection.getServerDescriptor().getGUID();
+    }
+
     public synchronized boolean isV1() {
         return !isV2();
     }
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -37,9 +37,13 @@
  */
 package io.cryostat.core.net;
 
+import java.io.ByteArrayOutputStream;
+import java.io.DataOutputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -68,7 +72,7 @@ import io.cryostat.core.templates.MergedTemplateService;
 import io.cryostat.core.templates.TemplateService;
 import io.cryostat.core.tui.ClientWriter;
 
-import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.codec.digest.DigestUtils;
 
 public class JFRConnection implements AutoCloseable {
 
@@ -154,34 +158,33 @@ public class JFRConnection implements AutoCloseable {
         }
     }
 
-    public synchronized int getJvmId() {
-        try {
-            if (!isConnected()) {
-                connect();
-            }
-            ObjectName runtimeBean = new ObjectName("java.lang:type=Runtime");
-            List<String> attributes =
-                    new ArrayList<>(
-                            Arrays.asList(
-                                    "ClassPath",
-                                    "Name",
-                                    "InputArguments",
-                                    "LibraryPath",
-                                    "VmVendor",
-                                    "VmVersion",
-                                    "StartTime"));
-            HashCodeBuilder jvmId = new HashCodeBuilder();
-
-            for (String attr : attributes) {
-                jvmId.append(
-                        this.rjmxConnection.getAttributeValue(
-                                new MRI(Type.ATTRIBUTE, runtimeBean, attr)));
-            }
-            return jvmId.hashCode();
-        } catch (Exception e) {
-            cw.println(e);
-            return 0;
+    public synchronized String getJvmId() throws Exception {
+        if (!isConnected()) {
+            connect();
         }
+        ObjectName runtimeBean = new ObjectName("java.lang:type=Runtime");
+        List<String> attrNames =
+                new ArrayList<>(
+                        Arrays.asList(
+                                "ClassPath",
+                                "Name",
+                                "InputArguments",
+                                "LibraryPath",
+                                "VmVendor",
+                                "VmVersion",
+                                "StartTime"));
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
+        DataOutputStream dos = new DataOutputStream(baos);
+
+        for (String attr : attrNames) {
+            dos.writeUTF(
+                    this.rjmxConnection
+                            .getAttributeValue(new MRI(Type.ATTRIBUTE, runtimeBean, attr))
+                            .toString());
+        }
+        byte[] hash = DigestUtils.sha256(baos.toByteArray());
+        return new String(Base64.getUrlEncoder().encode(hash), StandardCharsets.UTF_8).trim();
     }
 
     public synchronized boolean isV1() {
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -161,7 +161,9 @@ public class JFRConnection implements AutoCloseable {
         }
     }
 
-    public synchronized String getJvmId() throws AttributeNotFoundException, InstanceNotFoundException, MBeanException, ReflectionException, IOException {
+    public synchronized String getJvmId()
+            throws AttributeNotFoundException, InstanceNotFoundException, MBeanException,
+                    ReflectionException, IOException {
         if (!isConnected()) {
             connect();
         }
@@ -182,15 +184,10 @@ public class JFRConnection implements AutoCloseable {
                 Object attrObject =
                         this.rjmxConnection.getAttributeValue(
                                 new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
-
-                if (attrObject instanceof String || attrObject instanceof Long) {
-                    dos.writeUTF(attrObject.toString());
-                } else if (attrObject instanceof String[]) {
+                if (attrObject.getClass().isArray()) {
                     dos.writeUTF(Arrays.toString((String[]) attrObject));
                 } else {
-                    throw new IllegalArgumentException(
-                            String.format(
-                                    "Unexpected attribute: {%s} of type: {%s}", attr, attrObject.getClass()));
+                    dos.writeUTF(attrObject.toString());
                 }
             }
             byte[] hash = DigestUtils.sha256(baos.toByteArray());
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -47,6 +47,10 @@ import java.util.Base64;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.management.AttributeNotFoundException;
+import javax.management.InstanceNotFoundException;
+import javax.management.MBeanException;
+import javax.management.ReflectionException;
 import javax.management.remote.JMXServiceURL;
 
 import org.openjdk.jmc.rjmx.ConnectionException;
@@ -157,7 +161,7 @@ public class JFRConnection implements AutoCloseable {
         }
     }
 
-    public synchronized String getJvmId() throws Exception {
+    public synchronized String getJvmId() throws AttributeNotFoundException, InstanceNotFoundException, MBeanException, ReflectionException, IOException {
         if (!isConnected()) {
             connect();
         }
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -47,7 +47,6 @@ import java.util.Base64;
 import java.util.List;
 import java.util.stream.Collectors;
 
-import javax.management.ObjectName;
 import javax.management.remote.JMXServiceURL;
 
 import org.openjdk.jmc.rjmx.ConnectionException;
@@ -162,7 +161,6 @@ public class JFRConnection implements AutoCloseable {
         if (!isConnected()) {
             connect();
         }
-        ObjectName runtimeBean = new ObjectName("java.lang:type=Runtime");
         List<String> attrNames =
                 new ArrayList<>(
                         Arrays.asList(
@@ -177,10 +175,19 @@ public class JFRConnection implements AutoCloseable {
         try (ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
                 DataOutputStream dos = new DataOutputStream(baos)) {
             for (String attr : attrNames) {
-                dos.writeUTF(
-                        this.rjmxConnection
-                                .getAttributeValue(new MRI(Type.ATTRIBUTE, runtimeBean, attr))
-                                .toString());
+                Object attrObject =
+                        this.rjmxConnection.getAttributeValue(
+                                new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
+
+                if (attrObject instanceof String || attrObject instanceof Long) {
+                    dos.writeUTF(attrObject.toString());
+                } else if (attrObject instanceof String[]) {
+                    dos.writeUTF(Arrays.toString((String[]) attrObject));
+                } else {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Unexpected attribute class of %s", attrObject.getClass()));
+                }
             }
             byte[] hash = DigestUtils.sha256(baos.toByteArray());
             return new String(Base64.getUrlEncoder().encode(hash), StandardCharsets.UTF_8).trim();
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -39,9 +39,16 @@ package io.cryostat.core.net;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.management.AttributeNotFoundException;
+import javax.management.InstanceNotFoundException;
+import javax.management.MBeanException;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
+import javax.management.ReflectionException;
 import javax.management.remote.JMXServiceURL;
 
 import org.openjdk.jmc.rjmx.ConnectionException;
@@ -56,6 +63,8 @@ import org.openjdk.jmc.rjmx.internal.ServerDescriptor;
 import org.openjdk.jmc.rjmx.services.jfr.IFlightRecorderService;
 import org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderServiceFactory;
 import org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderServiceV2;
+import org.openjdk.jmc.rjmx.subscription.MRI;
+import org.openjdk.jmc.rjmx.subscription.MRI.Type;
 
 import io.cryostat.core.sys.Clock;
 import io.cryostat.core.sys.Environment;
@@ -63,6 +72,8 @@ import io.cryostat.core.sys.FileSystem;
 import io.cryostat.core.templates.MergedTemplateService;
 import io.cryostat.core.templates.TemplateService;
 import io.cryostat.core.tui.ClientWriter;
+
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 public class JFRConnection implements AutoCloseable {
 
@@ -148,8 +159,36 @@ public class JFRConnection implements AutoCloseable {
         }
     }
 
-    public synchronized String getGUID() {
-        return this.rjmxConnection.getServerDescriptor().getGUID();
+    public synchronized int getJvmId() {
+        try {
+            ObjectName runtimeBean = new ObjectName("java.lang:type=Runtime");
+            List<String> attributes =
+                    new ArrayList<>(
+                            Arrays.asList(
+                                    "ClassPath",
+                                    "Name",
+                                    "InputArguments",
+                                    "LibraryPath",
+                                    "VmVendor",
+                                    "VmVersion",
+                                    "StartTime"));
+            HashCodeBuilder jvmId = new HashCodeBuilder();
+
+            for (String attr : attributes) {
+                jvmId.append(
+                        this.rjmxConnection.getAttributeValue(
+                                new MRI(Type.ATTRIBUTE, runtimeBean, attr)));
+            }
+            return jvmId.hashCode();
+        } catch (MalformedObjectNameException
+                | IOException
+                | AttributeNotFoundException
+                | InstanceNotFoundException
+                | MBeanException
+                | ReflectionException e) {
+            cw.println(e);
+            return 0;
+        }
     }
 
     public synchronized boolean isV1() {
--- a/src/main/java/io/cryostat/core/net/JFRConnection.java
+++ b/src/main/java/io/cryostat/core/net/JFRConnection.java
@@ -186,7 +186,7 @@ public class JFRConnection implements AutoCloseable {
                         this.rjmxConnection.getAttributeValue(
                                 new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
                 if (attrObject.getClass().isArray()) {
-                    if (!attrObject.getClass().getComponentType().isPrimitive()) {
+                    if (attrObject.getClass().getComponentType().isPrimitive()) {
                         attrObject = castPrimitiveToObject(attrObject);
                     }
                     dos.writeUTF(Arrays.toString((Object[]) attrObject));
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -146,6 +157,38 @@ public synchronized int getPort() {
...
+                                new MRI(Type.ATTRIBUTE, runtimeBean, attr)));
+            }
+            return jvmId.hashCode();
+        } catch (MalformedObjectNameException
+                | IOException
+                | AttributeNotFoundException
+                | InstanceNotFoundException
+                | MBeanException
+                | ReflectionException e) {
+            cw.println(e);
```
jan-law, line 187: Where does this get printed to? 
I'm still looking for the cause of the NPE that sometimes occurs when I start an automated rule or perform a graphql query.

jan-law: Okay, thanks

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -146,6 +152,36 @@ public synchronized int getPort() {
         }
     }
 
+    public synchronized int getJvmId() {
+        try {
+            if (!isConnected()) {
+                connect();
+            }
```
jan-law, line 159: For the other helper functions like `getApproximateServerTime`, `getHost`, how is `rjmxConnection` initialized when Cryostat is connecting to target JVMs?

I found out `this.rjmxConnection` is sometimes null at this point when starting recordings with automated rules. If call `connect()` here, the jvmId returns successfully, but I think my connection is conflicting with `connection.getService().start(recordingOptions, enableEvents(...)` from `RecordingTargetHelper.java:526`, resulting in a `NoSuchElementException` whenever `enableEvents` is called:
```
SEVERE: Exception thrown
java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:148)
	at io.cryostat.recordings.RecordingTargetHelper.enableEvents(RecordingTargetHelper.java:526)
	at io.cryostat.recordings.RecordingTargetHelper.lambda$1(RecordingTargetHelper.java:168)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:156)
	at io.cryostat.recordings.RecordingTargetHelper.startRecording(RecordingTargetHelper.java:146)
	at io.cryostat.rules.RuleProcessor.lambda$6(RuleProcessor.java:275)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:156)
	at io.cryostat.rules.RuleProcessor.startRuleRecording(RuleProcessor.java:259)
	at io.cryostat.rules.RuleProcessor.activate(RuleProcessor.java:178)
	at io.cryostat.rules.RuleProcessor.lambda$2(RuleProcessor.java:137)
```

jan-law: Ah my bad, it turns out this was a side effect of https://github.com/cryostatio/cryostat-web/issues/433 - everything is working now.

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -146,6 +152,36 @@ public synchronized int getPort() {
...
+
+            for (String attr : attributes) {
+                jvmId.append(
+                        this.rjmxConnection.getAttributeValue(
+                                new MRI(Type.ATTRIBUTE, runtimeBean, attr)));
+            }
+            return jvmId.hashCode();
+        } catch (Exception e) {
+            cw.println(e);
+            return 0;
```
andrewazores, line 181: I think I'd prefer to re-throw than to return 0 here. This would swallow connection exceptions occurring because the target JVM has gone offline, for example, and we'd end up mapping all such JVMs to the same 0 ID.

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -146,6 +156,35 @@ public synchronized int getPort() {
...
+                                "ClassPath",
+                                "Name",
+                                "InputArguments",
+                                "LibraryPath",
+                                "VmVendor",
+                                "VmVersion",
+                                "StartTime"));
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
+        DataOutputStream dos = new DataOutputStream(baos);
```
andrewazores, line 176: Just for good measure, should these streams be closed? A try-with-resources block would be nice.

jan-law: Yes, thanks

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -146,6 +156,35 @@ public synchronized int getPort() {
...
+                                "StartTime"));
+
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
+                DataOutputStream dos = new DataOutputStream(baos)) {
+            for (String attr : attrNames) {
+                dos.writeUTF(
+                        this.rjmxConnection
+                                .getAttributeValue(new MRI(Type.ATTRIBUTE, runtimeBean, attr))
+                                .toString());
+            }
```
andrewazores, line 182: I just had a thought here - might it make sense to include the machine's IP address in the stream before hashing it? I don't think we can retrieve that in the same uniform way through the Runtime bean, but it can still be appended after all those attributes are written in.

In the case of long-lived deployments on bare metal the IP address is very likely to be stable long term, not changing during the target JVM lifecycle anyway. In the case of deployments on a container platform like OpenShift then each replica of a target service will have its own unique IP address since they'd be sitting behind a load balancer, so this would eliminate the narrow timing window possibility where multiple replicas come up at the exact same `StartTime` (since all other attributes would also be the same for replicas of a service). If the deployment is scaled down or a replica crahses then we should notice that from the target discovery mechanism and can use that to clear any cached computed IDs, so even if a replacement replica comes online later (restart or scale back up) and the same IP happens to be re-assigned, we can treat it as a "fresh" target since the caches have already been cleared.

jan-law: Sounds good. I'll add that in

andrewazores: I guess it's possible, especially on bare metal deployments, that there are actually multiple IP addresses for multiple network interfaces. I guess all of those should be put into the hash algo.

jan-law: Sorry, where are the machine IPs located? None of the mbeans seem to show networking info.

andrewazores: Shoot, sorry for the wild goose chase. I took a look and couldn't seem to find any way to access it either. I thought at the least we could grab it from the open JMX connection itself, but even that doesn't look like it'll work.

Forget about it for now. I'll keep thinking if there are any ways to do this, or other things we can add to the hash input.

jan-law: Okay no worries

maxcao13:
About this PR, I've been trying it out myself and the jvm id does not seem to be unique as referenced on June 2 here https://github.com/cryostatio/cryostat-web/issues/442#issuecomment-1144214316. I'm wondering if that's still the current state of this PR and the next step is to try and make them unique.

I still get something like this
```
localhost:9091: E-8K-XuIaFfjzCcnKS-4zb2ujb8oywXiz8yRv5fshWo=

service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi: xYHlk9oYqJrz3Li6FBN8y3y3r6fs8KQ_-muay4OZkBE=

cryostat:9091: RPYotmizzDwC8-nZRkxKUs5iULupqGKoyAw-bJcIh_w=
```

andrewazores:
I haven't tested this out myself before, but looking at the code it seems like it should work as expected. Can you try modifying it to also print out the values of each of those MBean attributes that are going into the hash function and determine what's causing the difference?

maxcao13:
For `service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi`
```
ClassPath: /app/resources:/app/classes:/app/libs/cryostat-core-2.13.0-SNAPSHOT.jar:/app/libs/common-7.1.1.jar:/app/libs/encoder-1.2.2.jar:/app/libs/flightrecorder-7.1.1.jar:/app/libs/flightrecorder.rules-7.1.1.jar:/app/libs/flightrecorder.rules.jdk-7.1.1.jar:/app/libs/nashorn-core-15.4.jar:/app/libs/asm-7.3.1.jar:/app/libs/asm-commons-7.3.1.jar:/app/libs/asm-analysis-7.3.1.jar:/app/libs/asm-tree-7.3.1.jar:/app/libs/asm-util-7.3.1.jar:/app/libs/openshift-client-5.4.1.jar:/app/libs/openshift-model-5.4.1.jar:/app/libs/kubernetes-model-common-5.4.1.jar:/app/libs/jackson-annotations-2.11.2.jar:/app/libs/openshift-model-operator-5.4.1.jar:/app/libs/openshift-model-operatorhub-5.4.1.jar:/app/libs/openshift-model-monitoring-5.4.1.jar:/app/libs/openshift-model-console-5.4.1.jar:/app/libs/kubernetes-client-5.4.1.jar:/app/libs/kubernetes-model-core-5.4.1.jar:/app/libs/kubernetes-model-rbac-5.4.1.jar:/app/libs/kubernetes-model-admissionregistration-5.4.1.jar:/app/libs/kubernetes-model-apps-5.4.1.jar:/app/libs/kubernetes-model-autoscaling-5.4.1.jar:/app/libs/kubernetes-model-apiextensions-5.4.1.jar:/app/libs/kubernetes-model-batch-5.4.1.jar:/app/libs/kubernetes-model-certificates-5.4.1.jar:/app/libs/kubernetes-model-coordination-5.4.1.jar:/app/libs/kubernetes-model-discovery-5.4.1.jar:/app/libs/kubernetes-model-events-5.4.1.jar:/app/libs/kubernetes-model-extensions-5.4.1.jar:/app/libs/kubernetes-model-flowcontrol-5.4.1.jar:/app/libs/kubernetes-model-networking-5.4.1.jar:/app/libs/kubernetes-model-metrics-5.4.1.jar:/app/libs/kubernetes-model-policy-5.4.1.jar:/app/libs/kubernetes-model-scheduling-5.4.1.jar:/app/libs/kubernetes-model-storageclass-5.4.1.jar:/app/libs/kubernetes-model-node-5.4.1.jar:/app/libs/okhttp-3.12.12.jar:/app/libs/okio-1.15.0.jar:/app/libs/logging-interceptor-3.12.12.jar:/app/libs/slf4j-api-1.7.30.jar:/app/libs/jackson-dataformat-yaml-2.11.2.jar:/app/libs/snakeyaml-1.26.jar:/app/libs/jackson-datatype-jsr310-2.11.2.jar:/app/libs/jackson-databind-2.11.2.jar:/app/libs/jackson-core-2.11.2.jar:/app/libs/zjsonpatch-0.3.0.jar:/app/libs/generex-1.0.2.jar:/app/libs/automaton-1.11-8.jar:/app/libs/dagger-2.34.1.jar:/app/libs/javax.inject-1.jar:/app/libs/commons-lang3-3.12.0.jar:/app/libs/commons-codec-1.15.jar:/app/libs/commons-io-2.8.0.jar:/app/libs/commons-validator-1.7.jar:/app/libs/commons-beanutils-1.9.4.jar:/app/libs/commons-digester-2.1.jar:/app/libs/commons-logging-1.2.jar:/app/libs/commons-collections-3.2.2.jar:/app/libs/httpclient-4.5.13.jar:/app/libs/httpcore-4.4.13.jar:/app/libs/vertx-web-4.2.5.jar:/app/libs/vertx-web-common-4.2.5.jar:/app/libs/vertx-auth-common-4.2.5.jar:/app/libs/vertx-bridge-common-4.2.5.jar:/app/libs/vertx-core-4.2.5.jar:/app/libs/netty-common-4.1.74.Final.jar:/app/libs/netty-buffer-4.1.74.Final.jar:/app/libs/netty-transport-4.1.74.Final.jar:/app/libs/netty-handler-4.1.74.Final.jar:/app/libs/netty-codec-4.1.74.Final.jar:/app/libs/netty-tcnative-classes-2.0.48.Final.jar:/app/libs/netty-handler-proxy-4.1.74.Final.jar:/app/libs/netty-codec-socks-4.1.74.Final.jar:/app/libs/netty-codec-http-4.1.74.Final.jar:/app/libs/netty-codec-http2-4.1.74.Final.jar:/app/libs/netty-resolver-4.1.74.Final.jar:/app/libs/netty-resolver-dns-4.1.74.Final.jar:/app/libs/netty-codec-dns-4.1.74.Final.jar:/app/libs/vertx-web-client-4.2.5.jar:/app/libs/vertx-web-graphql-4.2.5.jar:/app/libs/graphql-java-17.3.jar:/app/libs/java-dataloader-3.1.0.jar:/app/libs/antlr4-runtime-4.9.2.jar:/app/libs/reactive-streams-1.0.3.jar:/app/libs/graphql-java-extended-scalars-17.0.jar:/app/libs/nimbus-jose-jwt-9.16.1.jar:/app/libs/jcip-annotations-1.0-1.jar:/app/libs/bcprov-jdk15on-1.69.jar:/app/libs/slf4j-jdk14-1.7.30.jar:/app/libs/gson-2.8.9.jar:/app/libs/caffeine-3.0.1.jar:/app/libs/jsoup-1.14.2.jar:/clientlib/*
Name: 1@cryostat
InputArguments: [Ljava.lang.String;@3aa2bb
LibraryPath: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
VmVendor: Red Hat, Inc.
VmVersion: 17.0.3+7-LTS
StartTime: 1660668446705
Aug 16, 2022 4:47:33 PM io.cryostat.core.log.Logger info
INFO: JVM ID: aOqB4fGDj2qddKtPd-YePy5tOE4nki2whr-W_gfxVo0=
```

For `localhost:9091`
```
ClassPath: /app/resources:/app/classes:/app/libs/cryostat-core-2.13.0-SNAPSHOT.jar:/app/libs/common-7.1.1.jar:/app/libs/encoder-1.2.2.jar:/app/libs/flightrecorder-7.1.1.jar:/app/libs/flightrecorder.rules-7.1.1.jar:/app/libs/flightrecorder.rules.jdk-7.1.1.jar:/app/libs/nashorn-core-15.4.jar:/app/libs/asm-7.3.1.jar:/app/libs/asm-commons-7.3.1.jar:/app/libs/asm-analysis-7.3.1.jar:/app/libs/asm-tree-7.3.1.jar:/app/libs/asm-util-7.3.1.jar:/app/libs/openshift-client-5.4.1.jar:/app/libs/openshift-model-5.4.1.jar:/app/libs/kubernetes-model-common-5.4.1.jar:/app/libs/jackson-annotations-2.11.2.jar:/app/libs/openshift-model-operator-5.4.1.jar:/app/libs/openshift-model-operatorhub-5.4.1.jar:/app/libs/openshift-model-monitoring-5.4.1.jar:/app/libs/openshift-model-console-5.4.1.jar:/app/libs/kubernetes-client-5.4.1.jar:/app/libs/kubernetes-model-core-5.4.1.jar:/app/libs/kubernetes-model-rbac-5.4.1.jar:/app/libs/kubernetes-model-admissionregistration-5.4.1.jar:/app/libs/kubernetes-model-apps-5.4.1.jar:/app/libs/kubernetes-model-autoscaling-5.4.1.jar:/app/libs/kubernetes-model-apiextensions-5.4.1.jar:/app/libs/kubernetes-model-batch-5.4.1.jar:/app/libs/kubernetes-model-certificates-5.4.1.jar:/app/libs/kubernetes-model-coordination-5.4.1.jar:/app/libs/kubernetes-model-discovery-5.4.1.jar:/app/libs/kubernetes-model-events-5.4.1.jar:/app/libs/kubernetes-model-extensions-5.4.1.jar:/app/libs/kubernetes-model-flowcontrol-5.4.1.jar:/app/libs/kubernetes-model-networking-5.4.1.jar:/app/libs/kubernetes-model-metrics-5.4.1.jar:/app/libs/kubernetes-model-policy-5.4.1.jar:/app/libs/kubernetes-model-scheduling-5.4.1.jar:/app/libs/kubernetes-model-storageclass-5.4.1.jar:/app/libs/kubernetes-model-node-5.4.1.jar:/app/libs/okhttp-3.12.12.jar:/app/libs/okio-1.15.0.jar:/app/libs/logging-interceptor-3.12.12.jar:/app/libs/slf4j-api-1.7.30.jar:/app/libs/jackson-dataformat-yaml-2.11.2.jar:/app/libs/snakeyaml-1.26.jar:/app/libs/jackson-datatype-jsr310-2.11.2.jar:/app/libs/jackson-databind-2.11.2.jar:/app/libs/jackson-core-2.11.2.jar:/app/libs/zjsonpatch-0.3.0.jar:/app/libs/generex-1.0.2.jar:/app/libs/automaton-1.11-8.jar:/app/libs/dagger-2.34.1.jar:/app/libs/javax.inject-1.jar:/app/libs/commons-lang3-3.12.0.jar:/app/libs/commons-codec-1.15.jar:/app/libs/commons-io-2.8.0.jar:/app/libs/commons-validator-1.7.jar:/app/libs/commons-beanutils-1.9.4.jar:/app/libs/commons-digester-2.1.jar:/app/libs/commons-logging-1.2.jar:/app/libs/commons-collections-3.2.2.jar:/app/libs/httpclient-4.5.13.jar:/app/libs/httpcore-4.4.13.jar:/app/libs/vertx-web-4.2.5.jar:/app/libs/vertx-web-common-4.2.5.jar:/app/libs/vertx-auth-common-4.2.5.jar:/app/libs/vertx-bridge-common-4.2.5.jar:/app/libs/vertx-core-4.2.5.jar:/app/libs/netty-common-4.1.74.Final.jar:/app/libs/netty-buffer-4.1.74.Final.jar:/app/libs/netty-transport-4.1.74.Final.jar:/app/libs/netty-handler-4.1.74.Final.jar:/app/libs/netty-codec-4.1.74.Final.jar:/app/libs/netty-tcnative-classes-2.0.48.Final.jar:/app/libs/netty-handler-proxy-4.1.74.Final.jar:/app/libs/netty-codec-socks-4.1.74.Final.jar:/app/libs/netty-codec-http-4.1.74.Final.jar:/app/libs/netty-codec-http2-4.1.74.Final.jar:/app/libs/netty-resolver-4.1.74.Final.jar:/app/libs/netty-resolver-dns-4.1.74.Final.jar:/app/libs/netty-codec-dns-4.1.74.Final.jar:/app/libs/vertx-web-client-4.2.5.jar:/app/libs/vertx-web-graphql-4.2.5.jar:/app/libs/graphql-java-17.3.jar:/app/libs/java-dataloader-3.1.0.jar:/app/libs/antlr4-runtime-4.9.2.jar:/app/libs/reactive-streams-1.0.3.jar:/app/libs/graphql-java-extended-scalars-17.0.jar:/app/libs/nimbus-jose-jwt-9.16.1.jar:/app/libs/jcip-annotations-1.0-1.jar:/app/libs/bcprov-jdk15on-1.69.jar:/app/libs/slf4j-jdk14-1.7.30.jar:/app/libs/gson-2.8.9.jar:/app/libs/caffeine-3.0.1.jar:/app/libs/jsoup-1.14.2.jar:/clientlib/*
Name: 1@cryostat
InputArguments: [Ljava.lang.String;@76df25d5
LibraryPath: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
VmVendor: Red Hat, Inc.
VmVersion: 17.0.3+7-LTS
StartTime: 1660668446705
Aug 16, 2022 4:53:33 PM io.cryostat.core.log.Logger info
INFO: JVM ID: H7J6IXbyP8SyR5UJ-GAqnqFWqZIR9IbkCj9UzINS0Dc=
```
For `cryostat:9091`
```
ClassPath: /app/resources:/app/classes:/app/libs/cryostat-core-2.13.0-SNAPSHOT.jar:/app/libs/common-7.1.1.jar:/app/libs/encoder-1.2.2.jar:/app/libs/flightrecorder-7.1.1.jar:/app/libs/flightrecorder.rules-7.1.1.jar:/app/libs/flightrecorder.rules.jdk-7.1.1.jar:/app/libs/nashorn-core-15.4.jar:/app/libs/asm-7.3.1.jar:/app/libs/asm-commons-7.3.1.jar:/app/libs/asm-analysis-7.3.1.jar:/app/libs/asm-tree-7.3.1.jar:/app/libs/asm-util-7.3.1.jar:/app/libs/openshift-client-5.4.1.jar:/app/libs/openshift-model-5.4.1.jar:/app/libs/kubernetes-model-common-5.4.1.jar:/app/libs/jackson-annotations-2.11.2.jar:/app/libs/openshift-model-operator-5.4.1.jar:/app/libs/openshift-model-operatorhub-5.4.1.jar:/app/libs/openshift-model-monitoring-5.4.1.jar:/app/libs/openshift-model-console-5.4.1.jar:/app/libs/kubernetes-client-5.4.1.jar:/app/libs/kubernetes-model-core-5.4.1.jar:/app/libs/kubernetes-model-rbac-5.4.1.jar:/app/libs/kubernetes-model-admissionregistration-5.4.1.jar:/app/libs/kubernetes-model-apps-5.4.1.jar:/app/libs/kubernetes-model-autoscaling-5.4.1.jar:/app/libs/kubernetes-model-apiextensions-5.4.1.jar:/app/libs/kubernetes-model-batch-5.4.1.jar:/app/libs/kubernetes-model-certificates-5.4.1.jar:/app/libs/kubernetes-model-coordination-5.4.1.jar:/app/libs/kubernetes-model-discovery-5.4.1.jar:/app/libs/kubernetes-model-events-5.4.1.jar:/app/libs/kubernetes-model-extensions-5.4.1.jar:/app/libs/kubernetes-model-flowcontrol-5.4.1.jar:/app/libs/kubernetes-model-networking-5.4.1.jar:/app/libs/kubernetes-model-metrics-5.4.1.jar:/app/libs/kubernetes-model-policy-5.4.1.jar:/app/libs/kubernetes-model-scheduling-5.4.1.jar:/app/libs/kubernetes-model-storageclass-5.4.1.jar:/app/libs/kubernetes-model-node-5.4.1.jar:/app/libs/okhttp-3.12.12.jar:/app/libs/okio-1.15.0.jar:/app/libs/logging-interceptor-3.12.12.jar:/app/libs/slf4j-api-1.7.30.jar:/app/libs/jackson-dataformat-yaml-2.11.2.jar:/app/libs/snakeyaml-1.26.jar:/app/libs/jackson-datatype-jsr310-2.11.2.jar:/app/libs/jackson-databind-2.11.2.jar:/app/libs/jackson-core-2.11.2.jar:/app/libs/zjsonpatch-0.3.0.jar:/app/libs/generex-1.0.2.jar:/app/libs/automaton-1.11-8.jar:/app/libs/dagger-2.34.1.jar:/app/libs/javax.inject-1.jar:/app/libs/commons-lang3-3.12.0.jar:/app/libs/commons-codec-1.15.jar:/app/libs/commons-io-2.8.0.jar:/app/libs/commons-validator-1.7.jar:/app/libs/commons-beanutils-1.9.4.jar:/app/libs/commons-digester-2.1.jar:/app/libs/commons-logging-1.2.jar:/app/libs/commons-collections-3.2.2.jar:/app/libs/httpclient-4.5.13.jar:/app/libs/httpcore-4.4.13.jar:/app/libs/vertx-web-4.2.5.jar:/app/libs/vertx-web-common-4.2.5.jar:/app/libs/vertx-auth-common-4.2.5.jar:/app/libs/vertx-bridge-common-4.2.5.jar:/app/libs/vertx-core-4.2.5.jar:/app/libs/netty-common-4.1.74.Final.jar:/app/libs/netty-buffer-4.1.74.Final.jar:/app/libs/netty-transport-4.1.74.Final.jar:/app/libs/netty-handler-4.1.74.Final.jar:/app/libs/netty-codec-4.1.74.Final.jar:/app/libs/netty-tcnative-classes-2.0.48.Final.jar:/app/libs/netty-handler-proxy-4.1.74.Final.jar:/app/libs/netty-codec-socks-4.1.74.Final.jar:/app/libs/netty-codec-http-4.1.74.Final.jar:/app/libs/netty-codec-http2-4.1.74.Final.jar:/app/libs/netty-resolver-4.1.74.Final.jar:/app/libs/netty-resolver-dns-4.1.74.Final.jar:/app/libs/netty-codec-dns-4.1.74.Final.jar:/app/libs/vertx-web-client-4.2.5.jar:/app/libs/vertx-web-graphql-4.2.5.jar:/app/libs/graphql-java-17.3.jar:/app/libs/java-dataloader-3.1.0.jar:/app/libs/antlr4-runtime-4.9.2.jar:/app/libs/reactive-streams-1.0.3.jar:/app/libs/graphql-java-extended-scalars-17.0.jar:/app/libs/nimbus-jose-jwt-9.16.1.jar:/app/libs/jcip-annotations-1.0-1.jar:/app/libs/bcprov-jdk15on-1.69.jar:/app/libs/slf4j-jdk14-1.7.30.jar:/app/libs/gson-2.8.9.jar:/app/libs/caffeine-3.0.1.jar:/app/libs/jsoup-1.14.2.jar:/clientlib/*
Name: 1@cryostat
InputArguments: [Ljava.lang.String;@1a118620
LibraryPath: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
VmVendor: Red Hat, Inc.
VmVersion: 17.0.3+7-LTS
StartTime: 1660668446705
Aug 16, 2022 4:55:11 PM io.cryostat.core.log.Logger info
INFO: JVM ID: AbNwhyzn9rCWqb4xHGziu4HrZEXTRFneAfFPQ8vddYQ=
```

So it seems only the InputArguments are different.

maxcao13:
After actually printing the contents  of `InputArguments` using `Arrays.toString()`, this is what is different

For `service:jmx:rmi:///jndi/rmi://localhost:9091/jmxrmi`

```
[-XX:+CrashOnOutOfMemoryError, -Dcom.sun.management.jmxremote.port=9091, -Dcom.sun.management.jmxremote.rmi.port=9091, -Djavax.net.ssl.trustStore=/opt/cryostat.d/truststore.p12, -Djavax.net.ssl.trustStorePassword=mq-0LxTNIRWfpwlj_9ywiOdFscjWMzgR, -Dcom.sun.management.jmxremote.autodiscovery=true, -Dcom.sun.management.jmxremote.authenticate=false, -Dcom.sun.management.jmxremote.ssl=false, -Dcom.sun.management.jmxremote.registry.ssl=false]

```

For service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi
```
[-XX:+CrashOnOutOfMemoryError, -Dcom.sun.management.jmxremote.port=9091, -Dcom.sun.management.jmxremote.rmi.port=9091, -Djavax.net.ssl.trustStore=/opt/cryostat.d/truststore.p12, -Djavax.net.ssl.trustStorePassword=mq-0LxTNIRWfpwlj_9ywiOdFscjWMzgR, -Dcom.sun.management.jmxremote.autodiscovery=true, -Dcom.sun.management.jmxremote.authenticate=false, -Dcom.sun.management.jmxremote.ssl=false, -Dcom.sun.management.jmxremote.registry.ssl=false]
```
For `localhost:9091`
```
[-XX:+CrashOnOutOfMemoryError, -Dcom.sun.management.jmxremote.port=9091, -Dcom.sun.management.jmxremote.rmi.port=9091, -Djavax.net.ssl.trustStore=/opt/cryostat.d/truststore.p12, -Djavax.net.ssl.trustStorePassword=mq-0LxTNIRWfpwlj_9ywiOdFscjWMzgR, -Dcom.sun.management.jmxremote.autodiscovery=true, -Dcom.sun.management.jmxremote.authenticate=false, -Dcom.sun.management.jmxremote.ssl=false, -Dcom.sun.management.jmxremote.registry.ssl=false]

```


For `cryostat:9091`
```
[-XX:+CrashOnOutOfMemoryError, -Dcom.sun.management.jmxremote.port=9091, -Dcom.sun.management.jmxremote.rmi.port=9091, -Djavax.net.ssl.trustStore=/opt/cryostat.d/truststore.p12, -Djavax.net.ssl.trustStorePassword=mq-0LxTNIRWfpwlj_9ywiOdFscjWMzgR, -Dcom.sun.management.jmxremote.autodiscovery=true, -Dcom.sun.management.jmxremote.authenticate=false, -Dcom.sun.management.jmxremote.ssl=false, -Dcom.sun.management.jmxremote.registry.ssl=false]

```

maxcao13:
The `-Djavax.net.ssl.trustStorePassword=mq-0LxTNIRWfpwlj_9ywiOdFscjWMzgR` are different.

andrewazores:
That's freaky. These are all different URLs checked against the same single running instance of Cryostat?

maxcao13:
Yup, the same instance has been running and I've just been checking the log file for my prints.

andrewazores:
Wow, I have no explanation for that... that JVM system property should just be getting [set once at startup](https://github.com/cryostatio/cryostat/blob/74671b8b69174975d478ea83d79e8f895710e64b/src/main/extras/app/entrypoint.sh#L112).

Just saw the edit - phew.

andrewazores:
Well, `List.of("a").equals(List.of("a"))`, but `List.of("a") != List.of("a")`. The reference being different isn't surprising to me.

```java
                dos.writeUTF(
                        this.rjmxConnection
                                .getAttributeValue(new MRI(Type.ATTRIBUTE, runtimeBean, attr))
                                .toString());
```

This part of the algorithm to concatenate all the attributes before hashing looks like it might be what's breaking down, though. If the `.getAttributeValue().toString()` is returning `"[Ljava.lang.String;@76df25d5"` instead of `"[-XX: ... ]` then it makes sense why that's broken.

If the `Object` returned by `getAttributeValue` `instanceof String` then you can `writeUTF` it directly into the stream to be hashed. If it's an array or List (or any other Collection type I suppose) then we need to do something other than `toString()`.

maxcao13:
Makes sense, I'll make a special case then and it'll probably fix it. Thanks!

andrewazores:
Under the hood it probably doesn't bother doing that because the operating system holds that information anyway, ex.

```
$ cat /proc/766892/cmdline
/usr/lib/jvm/java-17-openjdk-17.0.4.0.8-1.fc36.x86_64/bin/java-classpath/home/andrew/mvnd-0.8.0-linux-amd64/mvn/lib/ext/mvnd-common-0.8.0.jar:/home/andrew/mvnd-0.8.0-linux-amd64/mvn/lib/ext/mvnd-agent-0.8.0.jar-javaagent:/home/andrew/mvnd-0.8.0-linux-amd64/mvn/lib/ext/mvnd-agent-0.8.0.jar-Dmvnd.home=/home/andrew/mvnd-0.8.0-linux-amd64-Dmaven.home=/home/andrew/mvnd-0.8.0-linux-amd64/mvn-Dmaven.conf=/home/andrew/mvnd-0.8.0-linux-amd64/mvn/conf-Dmvnd.java.home=/usr/lib/jvm/java-17-openjdk-17.0.4.0.8-1.fc36.x86_64-Dlogback.configurationFile=/home/andrew/mvnd-0.8.0-linux-amd64/conf/logback.xml-Dmvnd.id=ee94c68c-Dmvnd.daemonStorage=/home/andrew/.m2/mvnd/registry/0.8.0-Dmvnd.registry=/home/andrew/.m2/mvnd/registry/0.8.0/registry.bin-Dmvnd.socketFamily=inet-Dmvnd.home=/home/andrew/mvnd-0.8.0-linux-amd64-Djdk.java.options=--add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/sun.net.www.protocol.jar=ALL-UNNAMED --add-opens java.base/sun.nio.fs=ALL-UNNAMED-Dmvnd.noDaemon=false-Dmvnd.debug=false-Dmvnd.idleTimeout=3h-Dmvnd.keepAlive=100ms-Dmvnd.extClasspath=-Dmvnd.coreExtensions=-Dmvnd.enableAssertions=false-Dmvnd.expirationCheckDelay=10s-Dmvnd.duplicateDaemonGracePeriod=10s-Dmvnd.socketFamily=inetorg.mvndaemon.mvnd.common.MavenDaemon
```

so the Runtime MXBean implementation probably just reads it from a platform-specific source (this `/proc` way probably doesn't work for OpenJDK on Windows for example), and then just packages it up into a `List` and returns it. No need to bother with caching it because then it's taking up extra memory in the Java server process.

Especially for M(X)Beans where they can be invoked remotely, so the `List` needs to be serialized, sent over the network, and deserialized. Mapping that back to a consistent `List` reference is difficult or impossible, but as long as it's a conformant `List` implementation then the `.equals()` and `.hashCode()` are consistent and that should be enough.

maxcao13:
I pushed my commit onto https://github.com/jan-law/cryostat-core/tree/jvm-id but it doesn't seem to update this PR...

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -148,6 +161,43 @@ public synchronized int getPort() {
...
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
+                DataOutputStream dos = new DataOutputStream(baos)) {
+            for (String attr : attrNames) {
+                Object attrObject =
+                        this.rjmxConnection.getAttributeValue(
+                                new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
+
+                if (attrObject instanceof String || attrObject instanceof Long) {
+                    dos.writeUTF(attrObject.toString());
+                } else if (attrObject instanceof String[]) {
```
andrewazores, line 188: Maybe `attrObject.getClass().isArray()` here so that arrays of things other than `String` get handled the same? It's not critical since it's highly unlikely that the types of any of these bean attributes changes, but still.

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -148,6 +161,43 @@ public synchronized int getPort() {
...
+                    dos.writeUTF(Arrays.toString((String[]) attrObject));
+                } else {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Unexpected attribute: {%s} of type: {%s}", attr, attrObject.getClass()));
```
andrewazores, line 193: I think I'd rather be lenient than strict in this case. Probably better to have generated IDs that turn out to be non-unique for two URLs pointing at the same JVM than to simply refuse to generate any ID at all for such JVMs. Something like:

```
var attrO = getAttributeValue(runtimeBean, attr)
if isArray(attrO) dos.writeUTF(Arrays.toString(attrO)
else dos.writeUTF(attrO.toString())
```

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -148,6 +161,40 @@ public synchronized int getPort() {
...
+                                "StartTime"));
+
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
+                DataOutputStream dos = new DataOutputStream(baos)) {
+            for (String attr : attrNames) {
+                Object attrObject =
+                        this.rjmxConnection.getAttributeValue(
+                                new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
+                if (attrObject.getClass().isArray()) {
+                    dos.writeUTF(Arrays.toString((String[]) attrObject));
```
andrewazores, line 188: Ah, wait, the `(String[])` cast here might not be safe now. The `isArray()` check would pass for ex. `int[]`, but then this cast will fail. Hmm. You can use `attrObject.getClass().getComponentType()` to get the type of array element, though then you still need some way to cast it correctly.

maxcao13: Compiler allows me to cast like `(Object[])` actually, I think that would be sufficient right?

andrewazores: Try it with `jshell` perhaps:

```java
$ jshell
|  Welcome to JShell -- Version 17.0.4
|  For an introduction type: /help intro

jshell> var a = new int[] { 1, 2, 3}
a ==> int[3] { 1, 2, 3 }

jshell> var b = (Object[]) a
|  Error:
|  incompatible types: int[] cannot be converted to java.lang.Object[]
|  var b = (Object[]) a;
|                     ^

```

`int` is not a subclass of `Object`, so this failure makes sense.

maxcao13: Ah, I didn't realize that `int[]` or all arrays are instances of Object in Java, but indeed primitives are not.

src/main/java/io/cryostat/core/net/JFRConnection.java
```diff
@@ -148,6 +162,86 @@ public synchronized int getPort() {
...
+                DataOutputStream dos = new DataOutputStream(baos)) {
+            for (String attr : attrNames) {
+                Object attrObject =
+                        this.rjmxConnection.getAttributeValue(
+                                new MRI(Type.ATTRIBUTE, ConnectionToolkit.RUNTIME_BEAN_NAME, attr));
+                if (attrObject.getClass().isArray()) {
+                    if (attrObject.getClass().getComponentType().isPrimitive()) {
+                        attrObject = castPrimitiveToObject(attrObject);
+                    }
+                    dos.writeUTF(Arrays.toString((Object[]) attrObject));
```
andrewazores, line 192: I'm still suspicious that this cast works at runtime:

```
$ jshell
Obj|  Welcome to JShell -- Version 17.0.4
|  For an introduction type: /help intro
e
jshell> Object[] a = new boolean[] { false}
|  Error:
|  incompatible types: boolean[] cannot be converted to java.lang.Object[]
|  Object[] a = new boolean[] { false};
|               ^--------------------^
```

andrewazores, line 192: Do we really need to check and determine that the object is actually an array of primitives, then convert that to an array of boxed objects, just to stringify that? I think we can skip the `ArrayUtils.toObject()` calls if the `castPrimitiveToObject` is just refactored to something like `primitiveArrayToString` and calls/returns `Arrays.toString()` instead of `ArrayUtils.toObject()`. Does that make sense?

maxcao13: Yeah it makes sense, sorry lots of brainfarts, I will fix it in the morning.

andrewazores:
Looks good from a reading. I assume now that your test case results in the same hash ID when connecting with the various different URLs?

maxcao13:
Yup, everything looks fine from my point of view on this, other than if you have ideas for new things to add to the hash. Just trying to fix the bugs in 971 using this PR now.

andrewazores (approved):

PR #146, opened 2022-06-09
chore(pom): update to JDK17

### LINKED ISSUE #137: [Task] Update to JDK17

andrewazores:
Fixes #137

The `.mvn/jvm.config` file is needed for now as a workaround on https://github.com/diffplug/spotless/issues/834 . One Mockito static method has been removed so it was replaced with an equivalent. Everything else is pretty routine.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,7 +23,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-java@v2
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'adopt'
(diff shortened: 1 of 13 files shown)


tthvo (commented):
Would there be any issue if we build `cryostat-core` in local maven with this update while other parts are still using JDK 11?

Edit: Just saw JDK17 update issues on other repos. So, we should wait before fetching and building this new updates?

tthvo:
oh that makes sense thanks!

tthvo (approved):

maxcao13 (approved):

ebaron (approved):
Looks good. Would you mind filing an issue to remove the workaround once spotless is fixed?

Issue #148, opened 2022-06-14
[Task] Refactor getPredicateRuleFilter method from [cryostat] and [cryostat-reports] into -core


maxcao13:
The method was created once in [cryostat-reports] as part of [the /report path](https://github.com/cryostatio/cryostat-reports/blob/569036c038145ea94260e03c44080e78b89f8b5e/src/main/java/io/cryostat/reports/ReportResource.java#L130) before I realized that it needed to also be used in cryostat itself part of -core. It also needs to be removed from [cryostat](https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/util/RuleFilterParser.java), and refactored in cryostat [SubprocessReportGenerator](https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java)

PR #149, opened 2022-06-14
chore(util): extract RuleFilterParser utility

maxcao13:
Fixes part of #148 

Still need to refactor [cryostat] and [cryostat-reports] once my [PR is merged](https://github.com/cryostatio/cryostat-reports/pull/40)

### DIFF
--- a/src/main/java/io/cryostat/core/util/RuleFilterParser.java
+++ b/src/main/java/io/cryostat/core/util/RuleFilterParser.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package io.cryostat.core.util;
+
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.openjdk.jmc.flightrecorder.rules.IRule;
+import org.openjdk.jmc.flightrecorder.rules.RuleRegistry;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class RuleFilterParser {
+    private static final Set<String> RULE_IDS_SET =
+            RuleRegistry.getRules().stream().map(rule -> rule.getId()).collect(Collectors.toSet());
+
+    private static final Set<String> TOPIC_IDS_SET =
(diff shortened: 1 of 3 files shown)


maxcao13:
Yeah that makes more sense let's do that.

src/main/java/io/cryostat/core/util/RuleFilterParser.java
```diff
@@ -0,0 +1,79 @@
...
+            RuleRegistry.getRules().stream().map(rule -> rule.getId()).collect(Collectors.toSet());
+
+    private static final Set<String> TOPIC_IDS_SET =
+            RuleRegistry.getRules().stream()
+                    .map(rule -> rule.getTopic())
+                    .collect(Collectors.toSet());
+
+    private RuleFilterParser() {}
+
+    public static Predicate<IRule> getPredicateRuleFilter(String rawFilter) {
```
andrewazores, line 61: Rename method to just `parse()`?

src/main/java/io/cryostat/core/util/RuleFilterParser.java
```diff
@@ -0,0 +1,79 @@
...
+public class RuleFilterParser {
+    private static final Set<String> RULE_IDS_SET =
+            RuleRegistry.getRules().stream().map(rule -> rule.getId()).collect(Collectors.toSet());
+
+    private static final Set<String> TOPIC_IDS_SET =
+            RuleRegistry.getRules().stream()
+                    .map(rule -> rule.getTopic())
+                    .collect(Collectors.toSet());
+
+    private RuleFilterParser() {}
```
andrewazores, line 59: It might not be immediately useful now, but it could be a nice refactor to make this class actually instantiable and the `getPredicateRuleFilter`/`parse` method non-`static`. Just makes it easier to deal with for dependency injection and test mocking.

src/main/java/io/cryostat/core/util/RuleFilterParser.java
```diff
@@ -0,0 +1,79 @@
...
+public class RuleFilterParser {
+    private static final Set<String> RULE_IDS_SET =
+            RuleRegistry.getRules().stream().map(rule -> rule.getId()).collect(Collectors.toSet());
+
+    private static final Set<String> TOPIC_IDS_SET =
+            RuleRegistry.getRules().stream()
+                    .map(rule -> rule.getTopic())
+                    .collect(Collectors.toSet());
+
+    public RuleFilterParser() {}
```
andrewazores, line 59: No need to explicitly define this empty constructor

src/main/java/io/cryostat/core/util/RuleFilterParser.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+public class RuleFilterParser {
+    private static final Set<String> RULE_IDS_SET =
+            RuleRegistry.getRules().stream().map(rule -> rule.getId()).collect(Collectors.toSet());
+
+    private static final Set<String> TOPIC_IDS_SET =
+            RuleRegistry.getRules().stream()
+                    .map(rule -> rule.getTopic())
+                    .collect(Collectors.toSet());
```
andrewazores, line 57: If you make these non-`static` as well, you can add package-private constructors that let you define the set of rule IDs and topics. The public constructor can take no args and default to assigning these computed values taken from the `RuleRegistry`. Then you can add unit tests for this class which don't depend on the real `RuleRegistry` and actual JMC `IRule` set, so you can isolate the predicate matching behaviour.

maxcao13: That makes a lot of sense right, gotcha.

src/main/java/io/cryostat/core/util/RuleFilterParser.java
```diff
@@ -48,24 +48,36 @@
 import org.apache.commons.lang3.StringUtils;
 
 public class RuleFilterParser {
-    private static final Set<String> RULE_IDS_SET =
-            RuleRegistry.getRules().stream().map(rule -> rule.getId()).collect(Collectors.toSet());
+    private final Set<String> RULE_IDS_SET;
+    private final Set<String> TOPIC_IDS_SET;
```
andrewazores, line 52: Nice. Now that theses are instance members rather than `static`, the naming convention should be ex. `topicIds` not `TOPIC_IDS`. I would take the `set` out of the name, it's redundant info since that's already the type declaration.

src/test/java/io/cryostat/core/util/RuleFilterParserTest.java
```diff
@@ -0,0 +1,170 @@
...
+        topicSet = Set.of("Topic0", "Topic1", "Topic2");
+
+        Mockito.when(iRules.stream()).thenReturn(Stream.of(rule0, rule1, rule2, rule3, rule4));
+
+        parser = new RuleFilterParser(ruleSet, topicSet);
+    }
+
+    @Test
+    void shouldAcceptAllRules() {
+        String rawFilter = "Rule0,Rule1, rule4,     rule2, rule3   ";
```
andrewazores, line 89: Could you add a test where the filter has mixed rules and topics?

src/test/java/io/cryostat/core/util/RuleFilterParserTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        Predicate<IRule> result = parser.parse(rawFilter);
+        Collection<IRule> rules = iRules.stream().filter(result).collect(Collectors.toList());
+
+        MatcherAssert.assertThat(result, Matchers.notNullValue());
+        MatcherAssert.assertThat(rules, Matchers.hasSize(5));
+    }
+
+    @Test
+    void shouldAcceptAllWhenFilterBlank() {
```
andrewazores, line 125: You could combine this test with the last one using `@ParameterizedTest @NullAndEmptySource`. It's not much of a savings, and the test intent is already crystal clear the way you've written it, so this is not really a suggestion to change. Just a neat trick to take a look at. This one taken from Cryostat's `RuleTest`.

```java

    @ParameterizedTest
    @NullAndEmptySource
    void shouldThrowOnBlankName(String s) {
        IllegalArgumentException ex =
                Assertions.assertThrows(
                        IllegalArgumentException.class,
                        () -> {
                            builder.name(s)
                                    .matchExpression(MATCH_EXPRESSION)
                                    .eventSpecifier(EVENT_SPECIFIER)
                                    .build();
                        });
        MatcherAssert.assertThat(
                ex.getMessage(),
                Matchers.containsString("\"name\" cannot be blank, was \"" + s + "\""));
    }
```

andrewazores (approved):

PR #152, opened 2022-08-08
feat(discovery): port discovery types from cryostat

andrewazores:
Related to cryostatio/cryostat#940


andrewazores:
Closing. This should go into an `-api` package instead.

PR #153, opened 2022-08-11
chore(spotbugs): update spotbugs dependency from v4.2.3 to v4.7.1

### LINKED ISSUE #145: [Task] Update spotbugs version
Currently using spotbugs version 4.2.3, but 4.7.0 is out. This version should be bumped and any new bugs found fixed accordingly.

maxcao13:
Fixes #145 

Thoughts?


src/main/java/io/cryostat/core/agent/IllegalRuntimeException.java
```diff
@@ -0,0 +1,48 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package io.cryostat.core.agent;
+
+public class IllegalRuntimeException extends RuntimeException {
```
andrewazores, line 40: What is the "meaning" of this exception type supposed to be?

andrewazores: Is [`IllegalStateException`](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html) close enough to the same thing?

andrewazores: Or are there perhaps other existing subclasses of `RuntimeException` that are more suitable?

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/RuntimeException.html

maxcao13: Doesn't seem so after skimming through them, `IllegalStateException` makes the most sense to me.

andrewazores (approved):

Issue #95, opened 2021-08-03
[Task] Update to JMC 8


andrewazores:
JMC 8 (8.0.1 now) is released and available on Maven Central:

https://search.maven.org/search?q=org.openjdk.jmc

andrewazores:
Fixed in #201

PR #154, opened 2022-08-24
Fix NPE

roberttoyonaga:
### Summary
Fixes a NPE encountered when attempting to connect with a non-hotspot VM.

### Problem Description
See error stack trace [here](https://gist.github.com/roberttoyonaga/9149882f08a4ce031fa8aea3c03b516e).

NPE occurs because `FlightRecorderServiceV2.connection` is not initialized before use. This can happen in the `FlightRecorderServiceV2` constructor when `isDynamicFlightRecorderSupported(IConnectionHandle)` returns `false` because it calls `ConnectionToolkit.isHotSpot(IConnectionHandle)` which returns `false` for native images.  This allows `isFlightRecorderDisabled(IConnectionHandle)` to be evaluated which results in the stack trace linked above. 

### Solution
Initialize `FlightRecorderServiceV2.connection` earlier.

### DIFF
--- a/src/main/java/org/openjdk/jmc/rjmx/services/jfr/internal/FlightRecorderServiceV2.java
+++ b/src/main/java/org/openjdk/jmc/rjmx/services/jfr/internal/FlightRecorderServiceV2.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
- * 
+ *
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * The contents of this file are subject to the terms of either the Universal Permissive License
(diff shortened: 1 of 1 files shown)


andrewazores:
That's occurring in the forked JMC code that we're carrying for downstream build purposes, so this means that at some point JMC itself was impacted by the same bug too. Could you verify if the same NPE occurs when you try to connect to your NI JMX server using JMC?

If JMC is affected then the best flow would be to work with them to fix it upstream, and then we pull that commit back down here, possibly as part of #95.

roberttoyonaga:
> That's occurring in the forked JMC code that we're carrying for downstream build purposes, so this means that at some point JMC itself was impacted by the same bug too. Could you verify if the same NPE occurs when you try to connect to your NI JMX server using JMC?
> 
> If JMC is affected then the best flow would be to work with them to fix it upstream, and then we pull that commit back down here, possibly as part of #95. If JMC is not affected (ie this bug is already fixed upstream) then we can accept this minor patch for now, but it would be superseded by the upstream fix when #95 is finished.

Ok I see. Yup I'll give it a try with JMC too!

roberttoyonaga:
It seems as though I can connect to the NI mbean server and invoke operations on the FlightRecorderMXBean with JMC 8.01. So maybe the problem has been fixed upstream already.

roberttoyonaga:
Yup, it looks like the problem has been fixed: https://github.com/openjdk/jmc/blob/master/application/org.openjdk.jmc.rjmx.services.jfr/src/main/java/org/openjdk/jmc/rjmx/services/jfr/internal/FlightRecorderServiceV2.java#L114

andrewazores:
https://github.com/cryostatio/cryostat-core/issues/155

PR #171, opened 2022-12-23
chore(jfrconnection): refactor interface to allow other impls

andrewazores:
Related to https://github.com/cryostatio/cryostat/issues/1310

Refactoring the `JFRConnection` class into an interface. The original implementation is directly copied to `JFRJMXConnection` implementation. This makes it cleaner for the Cryostat application to stub out an HTTP-based `JFRConnection` to communicate with the Cryostat Agent.


maxcao13 (approved):
Looks good and makes sense to me.

PR #207, opened 2023-04-12
fix(fs): close directory file descriptor

andrewazores:
Related to https://github.com/cryostatio/cryostat/issues/1444

The FileSystem wrapper around `java.nio.file.Files` fails to close the `Stream<Path>` it opens when listing directory contents. This class is dependency-injected in Cryostat (for testing mock ease) and used in several frequently called codepaths. This leaking fd adds up quite quickly as Cryostat runs and serves requests.


tthvo (approved):
Looks good! Fixed the issue!

Issue #238, opened 2023-06-20
[Bug] JMC probe template validation error


andrewazores:
Using the following XML probe template definition, which I generated using the JMC wizard:

```xml
<jfragent>
    <config>
        <classprefix>__JFREvent</classprefix>
        <allowtostring>false</allowtostring>
        <allowconverter>false</allowconverter>
    </config>
    <events>
        <event id="io.cryostat.quarkus.demo.event1">
            <label>Cryostat Quarkus Demo Event 1</label>
            <class>jakarta.persistence.EntityManager</class>
            <stacktrace>true</stacktrace>
            <rethrow>true</rethrow>
            <location>WRAP</location>
            <method>
                <name>find</name>
                <descriptor>(Ljava/lang/Class;Ljava/lang/Object;)Ljava/lang/Object;</descriptor>
                <parameters>
                    <parameter index="0">
                        <name>klazz</name>
                        <contenttype>Class</contenttype>
                    </parameter>
                    <parameter index="1">
                        <name>key</name>
                        <contenttype>None</contenttype>
                    </parameter>
                </parameters>
            </method>
        </event>
    </events>
</jfragent>
```

I get the following stack trace when attempting to upload it to Cryostat on the Events view:

```
Jun 20, 2023 8:28:49 PM io.cryostat.core.log.Logger error
SEVERE: cvc-enumeration-valid: Value 'CLASS' is not facet-valid with respect to enumeration '[None, Bytes, Timestamp, Millis, Nanos, Ticks, Address, OSThread, JavaThread, StackTrace, Class, Percentage]'. It must be a value from the enumeration.
Jun 20, 2023 8:28:49 PM io.cryostat.core.log.Logger warn
WARNING: HTTP 400: cvc-enumeration-valid: Value 'CLASS' is not facet-valid with respect to enumeration '[None, Bytes, Timestamp, Millis, Nanos, Ticks, Address, OSThread, JavaThread, StackTrace, Class, Percentage]'. It must be a value from the enumeration.
io.vertx.ext.web.handler.HttpException: Bad Request
Caused by: io.cryostat.net.web.http.api.v2.ApiException: cvc-enumeration-valid: Value 'CLASS' is not facet-valid with respect to enumeration '[None, Bytes, Timestamp, Millis, Nanos, Ticks, Address, OSThread, JavaThread, StackTrace, Class, Percentage]'. It must be a value from the enumeration.
[...]

PR #241, opened 2023-06-22
chore(jmc-agent): clean up probe template utility

### LINKED ISSUE #239: [Bug] Cannot upload JMC Agent probe templates after a failure
```
Jun 20, 2023 8:32:28 PM io.cryostat.core.log.Logger error
SEVERE: new_file.xml
Jun 20, 2023 8:32:28 PM io.cryostat.core.log.Logger error
SEVERE: HTTP 500: new_file.xml
io.vertx.ext.web.handler.HttpException: Internal Server Error
Caused by: io.cryostat.net.web.http.api.v2.ApiException: new_file.xml
	at io.cryostat.net.web.http.api.v2.ProbeTemplateUploadHandler.handle(ProbeTemplateUploadHandler.java:161)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:108)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:71)
	at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
	at io.vertx.core.impl.ContextBase.lambda$null$0(ContextBase.java:137)
	at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:264)
	at io.vertx.core.impl.ContextBase.lambda$executeBlocking$1(ContextBase.java:135)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.nio.file.FileAlreadyExistsException: new_file.xml
	at io.cryostat.core.agent.LocalProbeTemplateService.addTemplate(LocalProbeTemplateService.java:98)
	at io.cryostat.net.web.http.api.v2.ProbeTemplateUploadHandler.handle(ProbeTemplateUploadHandler.java:136)
	... 11 more
Jun 20, 2023 8:32:28 PM org.slf4j.impl.JDK14LoggerAdapter fillCallerData
SEVERE: 10.89.1.31 - - [Tue, 20 Jun 2023 20:32:28 GMT] 11ms "POST /api/v2/probes/jmc-agent-probe.xml HTTP/1.1" 500 96 bytes "http://localhost:8181/events?agentTab=agent-template" "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0"
```

See #238 . After failing to upload a template once, a retry with the file corrected fails with the above. Once in this state it doesn't appear to be recoverable - all upload attempts will fail this way.

tthvo:
Fixes #239 

Changes include:

- `addTemplate` should check if template file exists before validating.
  - Add more contexts in case file-exist error occurs.
  - ProbeTemplate should correctly set its name.
- `addTemplate` should return the `ProbeTemplate` object so that the caller can directly call `serialize` on it.
- Input streams should properly close using the custom `trulyClose`.
- Rename methods to match returned values.

### DIFF
--- a/src/main/java/io/cryostat/core/agent/LocalProbeTemplateService.java
+++ b/src/main/java/io/cryostat/core/agent/LocalProbeTemplateService.java
@@ -87,18 +87,19 @@ public class LocalProbeTemplateService implements ProbeTemplateService {
         }
     }
 
-    public void addTemplate(InputStream inputStream, String filename)
+    public ProbeTemplate addTemplate(InputStream inputStream, String filename)
             throws FileAlreadyExistsException, IOException, SAXException {
+        Path path = fs.pathOf(env.getEnv(TEMPLATE_PATH), filename);
+        if (fs.exists(path)) {
+            throw new FileAlreadyExistsException(filename);
+        }
         try (inputStream) {
             ProbeTemplate template = new ProbeTemplate();
             template.setFileName(filename);
             // If validation fails this will throw a ProbeValidationException with details
             template.deserialize(inputStream);
-            Path path = fs.pathOf(env.getEnv(TEMPLATE_PATH), filename);
-            if (fs.exists(path)) {
-                throw new FileAlreadyExistsException(template.getFileName());
-            }
             fs.writeString(path, template.serialize());
+            return template;
         }
     }
 
@@ -138,8 +139,8 @@ public class LocalProbeTemplateService implements ProbeTemplateService {
                         Path fileName = path.getFileName();
                         if (fileName != null) {
                             ProbeTemplate template = new ProbeTemplate();
-                            template.deserialize(stream);
                             template.setFileName(fileName.toString());
+                            template.deserialize(stream);
                             templates.add(template);
                         }
                     }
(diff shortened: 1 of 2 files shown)


tthvo:
I found that the template file is actually put into storage despite the API returning 400. This causes duplicate upload error when you try to upload the valid template with the same name.

After some digging, looks like validation (i.e. for the template specified in #238) is behaving differently between:

- Uploading templates (i.e. success)
https://github.com/cryostatio/cryostat-core/blob/f386e3920e388500eda2fc0e0d0f5de899de8ff7/src/main/java/io/cryostat/core/agent/LocalProbeTemplateService.java#L90

- Reading templates (i.e. failure in both methods below)
https://github.com/cryostatio/cryostat-core/blob/f386e3920e388500eda2fc0e0d0f5de899de8ff7/src/main/java/io/cryostat/core/agent/LocalProbeTemplateService.java#L111
https://github.com/cryostatio/cryostat-core/blob/f386e3920e388500eda2fc0e0d0f5de899de8ff7/src/main/java/io/cryostat/core/agent/LocalProbeTemplateService.java#L130

With this PR changes, Cryostat will only call `addTemplate` when uploading. Hence, in the UI, upload will succeed so that attempts to send duplicate filename will fail with message about file already existing. This closes the issue in #239.

However, reloading the UI page will show the same issue in #238.

maxcao13 (approved):
Testing with this and https://github.com/cryostatio/cryostat/pull/1563 seems to fix the issues!

andrewazores (approved):

PR #245, opened 2023-06-28
feat(svc): add recording start method using event template name and type

andrewazores:
- chore(svc): extract EventOptionsBuilder from Cryostat application
- chore(svc): add recording start method using event template name and type

Related to https://github.com/cryostatio/cryostat-agent/issues/162
Related to https://github.com/cryostatio/cryostat/issues/1578

This extracts the `EventOptionsBuilder` from the Cryostat application into `-core`, then augments the JMC `IFlightRecorderService` by introducing a subinterface `CryostatFlightRecorderService` that allows for additional/higher-level operations, such as dynamically starting a recording using an event template name+type rather than the JMC `IConstrainedMap`. This will support starting recordings by network request by asking the Agent to simply use a `.jfc` file available to it locally, rather than needing to reserialize the settings defined by that `.jfc` and transmit that over the network, where those settings would be originally retrieved by querying the raw `.jfc` form from the Agent and parsing it to create the `IConstrainedMap` model.

~~A further enhancement to come later will be to allow supplying the entire XML (`.jfc`) `Document` on the recording creation request. The `CryostatFlightRecorderService` implementation will either convert this to the `IConstrainedMap` server-side, or pass it as-is to the Agent for conversion, so that the XML content of Custom event templates does not need to be reserialized between XML, POJO, and possibly JSON in several stages along the flow.~~

Some serialization-related classes are also extracted (and minor refactorings done) from the Cryostat application into -core here, since they will be useful for facilitating (de)serialization of these structures for communication between Cryostat and the Agent.

### DIFF
--- a/src/main/java/io/cryostat/core/EventOptionsBuilder.java
+++ b/src/main/java/io/cryostat/core/EventOptionsBuilder.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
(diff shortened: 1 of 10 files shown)


maxcao13 (approved):
Makes sense to me, looks good!

src/main/java/io/cryostat/core/EventOptionsBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+    public EventOptionsBuilder(ClientWriter cw, JFRConnection connection, Supplier<Boolean> v2)
+            throws ServiceNotAvailableException, IOException,
+                    org.openjdk.jmc.rjmx.services.jfr.FlightRecorderException {
+        this.isV2 = v2.get();
+        this.map = connection.getService().getDefaultEventOptions().emptyWithSameConstraints();
+        knownTypes = new HashMap<>();
+        eventIds = new HashMap<>();
+
+        if (!isV2) {
+            cw.println("Flight Recorder V1 is not supported");
```
tthvo, line 78: Should this be `Flight Recorder V2 is not supported`?

src/main/java/io/cryostat/core/net/JmxFlightRecorderService.java
```diff
@@ -0,0 +1,304 @@
...
+        return builder.build();
+    }
+
+    private TemplateType getPreferredTemplateType(
+            JFRConnection connection, String templateName, TemplateType templateType)
+            throws io.cryostat.core.FlightRecorderException {
+        if (templateType != null) {
+            return templateType;
+        }
+        if (templateName.equals("ALL")) {
```
tthvo, line 281: Wondering if we still need this check since its only called in `enableEvents` and we already have:

```java
if (templateName.equals("ALL")) {
    return enableAllEvents();
}
```

tthvo (commented):
Looks good! I just have some small questions above.

src/main/java/io/cryostat/core/EventOptionsBuilder.java
andrewazores: I think it's right - only JFR V2 is supported, not V1 (which is the old pre-OpenJDK opensourced implementation), so this message is emitted if the connection is not V2.

tthvo: Ahhh okayy I got a little confused there^^Thanks!

tthvo (approved):

PR #253, opened 2023-08-03
chore(license): change project license to Apache-2.0

andrewazores:


ebaron (commented):
Looks good, what do you think about keeping a copy of the UPL under a file like `JMC_LICENSE` or something?

ebaron (approved):

Issue #221, opened 2023-05-25
[Task] PR CI should test `-core` changes in a Cryostat container image against existing itests


andrewazores:
Currently, when a PR is opened, only `-core`'s own unit tests are run.

It would be great if, after that passes, the built `-core` artifact is then used to build a Cryostat container image, and that container image run through the Cryostat integration tests. This way any changes to `-core` are automatically tested for regressions in the primary consumer of the `-core` library.

`-agent` and `-reports` also consume `-core`, so they could also be tested against, but the Cryostat server is the most important one to test.

PR #266, opened 2023-09-29
feat(ci): test -core's build in Cryostat image

mwangggg:
fixes: #221


mwangggg:
demo: [run](https://github.com/mwangggg/cryostat-core/actions/runs/6353883050)

mwangggg:
demo: [run](https://github.com/mwangggg/cryostat-core/actions/runs/6354503978)

.github/workflows/cryostat-IT.yml
```diff
@@ -0,0 +1,119 @@
...
+        required: false
+        type: boolean
+        default: false
+    outputs:
+      image-version:
+        description: the Cryostat application version that will be built
+        value: ${{ jobs.get-pom-properties.outputs.image-version }}
+
+permissions:
+  contents: read
```
tthvo, line 17: These permissions need revising. I think we don't need `contents: read` or `packages:read` here.

tthvo: Interestingly, artitifact-related actions are granted an "undocumented token": https://github.com/actions/upload-artifact/issues/197

tthvo: > These permissions need revising. I think we don't need contents: read or packages:read here.

Could u help check?

.github/workflows/cryostat-IT.yml
```diff
@@ -0,0 +1,119 @@
...
+    - name: Install qemu
+      if: ${{ inputs.build-arch != 'amd64' }}
+      continue-on-error: true
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-user-static
+    - uses: actions/checkout@v4
+      if: always()
+      with:
+        repository: cryostatio/cryostat
```
tthvo, line 50:
```yaml
with:
  owner: ${{ github.repository_owner }}
  name: cryostat
  ...
```

We should use the owner annotation here?

.github/workflows/cryostat-IT.yml
```diff
@@ -0,0 +1,119 @@
...
+      uses: s4u/maven-settings-action@v2
+      with:
+        servers: '[{"id": "github", "username": "dummy", "password": "${env.GITHUB_TOKEN_REF}"}]'
+        githubServer: false
+    - uses: skjolber/maven-cache-github-action@v1
+      with:
+        step: restore
+    - run: git submodule init
+    - run: git submodule update
+    - run: mvn -B -U -Dbuild.arch=${{ inputs.build-arch }} clean package
```
tthvo, line 68: Does this image use the `-core` version built from `push/PR` commit?

tthvo: I think what we can do is cache the artifact from `build` job in `.github/workflows/ci.yaml`. Then, restore it here and do an offline build here.

.github/workflows/cryostat-IT.yml
tthvo: And else where.

tthvo (commented):
Nice job! Just some questions. Also, why not explore pushing the test image to `ghcr.io` instead of saving as an aritifacts?

This way, we can output the test image + rerun tests without building too.

.github/workflows/cryostat-IT.yml
tthvo: Will need to patch the `pom.xml` on cryostat side tho.

.github/workflows/cryostat-IT.yml
mwangggg: I essentially copied this from the cryostat repo, so the `ci-build-image.yml` file there should also be revised/any changes made here can be copied over there too?

tthvo: Yess! Exactly! Pls do so!

mwangggg: [run](https://github.com/mwangggg/cryostat-core/actions/runs/6356474255/job/17266197558) shows the permissions aren't needed... but I'm not certain if it's because it's my forked repo

tthvo: I just realized the default restrictive token already has these permissions:

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

We can remove it here.

mwangggg:
demo: https://github.com/mwangggg/cryostat-core/actions/runs/6381050918

.github/workflows/cryostat-IT.yml
mwangggg: there's no option for an "owner" for actions/checkout@v4

mwangggg: using `repository: ${{ github.repository_owner }}/cryostat` instead

.github/workflows/ci.yaml
```diff
@@ -16,19 +16,99 @@ on:
...
+      matrix:
+        arch: [amd64, arm64]
+    uses: ./.github/workflows/cryostat-IT.yml
+    with:
+      build-arch: ${{ matrix.arch }}
+      skip-itests: ${{ matrix.arch != 'amd64' }}
+  
+  push-to-ghcr:
+    runs-on: ubuntu-latest
+    needs: [cryostat-test]
```
tthvo, line 50: How about building the image and push it to `ghcr.io` first? Then, pull it and use it the integration test?

.github/workflows/cryostat-IT.yml
```diff
@@ -0,0 +1,121 @@
...
+  get-pom-properties:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        repository: ${{ github.repository_owner }}/cryostat
+        ref: main
+    - uses: Mudlet/xmlstarlet-action@master
+      with:
+        args: ed -P -L -u '//_:properties[_:io.cryostat.core.version="2.23.0"]/_:io.cryostat.core.version' -v "2.24.0-SNAPSHOT" pom.xml
```
tthvo, line 26: Can we remove `2.23.0` and pass in `2.24.0-SNAPSHOT` as workflow argument?

This way, we don't have to go update these versions every release.

.github/workflows/cryostat-IT.yml
```diff
@@ -0,0 +1,121 @@
...
+    needs: [get-pom-properties]
+    steps:
+    - name: Install qemu
+      if: ${{ inputs.build-arch != 'amd64' }}
+      continue-on-error: true
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-user-static
+    - uses: actions/checkout@v4
+      if: always()
```
tthvo, line 47: Remove? We don't really need it since it will run anyway.

.github/workflows/ci.yaml
```diff
@@ -16,19 +16,99 @@ on:
...
-        java-version: '11'
+        java-version: '17'
         distribution: 'temurin'
     - uses: skjolber/maven-cache-github-action@v1
       with:
         step: restore
     - run: mvn -B -U clean verify
+    - uses: actions/upload-artifact@v3
+      with:
+        name: cryostat-core
```
tthvo, line 32: We will need to add some unique identifier here. Otherwise, concurrent runs will overwrite each other.

.github/workflows/cryostat-IT.yml
```diff
@@ -0,0 +1,121 @@
...
+        servers: '[{"id": "github", "username": "dummy", "password": "${env.GITHUB_TOKEN_REF}"}]'
+        githubServer: false
+    - uses: skjolber/maven-cache-github-action@v1
+      with:
+        step: restore
+    - run: git submodule init
+    - run: git submodule update
+    - uses: actions/download-artifact@v3
+      with:
+        name: cryostat-core
```
tthvo, line 69: I am not sure if this worked. This will download the `-core` artifact in local repository but not in local maven repo. Even so, not sure if maven knows about it when manually unpacking this artifact into local maven repo.

tthvo: We will need to cache the entire local maven repo when building and saving `-core` artifact. Can we use https://github.com/actions/setup-java#caching-packages-dependencies?

mwangggg: isn't  `skjolber/maven-cache-github-action@v1` caching the maven dependencies?

aali309 (commented):
Looks good to me

.github/workflows/cryostat-build.yml
```diff
@@ -0,0 +1,84 @@
...
+  get-pom-properties:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        repository: ${{ github.repository_owner }}/cryostat
+        ref: main
+    - uses: Mudlet/xmlstarlet-action@master
+      with:
+        args: ed -P -L -u '//_:properties[_:io.cryostat.core.version=${{ env.CURRENT_VERSION }}]/_:io.cryostat.core.version' -v ${{ env.TEST_VERSION }} pom.xml
```
andrewazores, line 30: You can actually do this even more simply, without modifying the `pom.xml`:

```bash
$ ./mvnw -Dio.cryostat.core.version=${{ env.TEST_VERSION}} clean package
```

andrewazores: This way you also don't need the `CURRENT_VERSION` env var because you're letting Maven directly override the property value instead of doing a text-based substitution.

.github/workflows/ci.yaml
```diff
@@ -14,21 +18,141 @@ on:
       - v[0-9]+.[0-9]+
       - cryostat-v[0-9]+.[0-9]+
 
+env:
+  TEST_VERSION: 2.25.0-SNAPSHOT
```
andrewazores, line 22: I think rather than hardcoding this value, it should just be read from the `pom.xml`. It could even be read from the PR's version of the `pom.xml` in case the PR changed the version string. This could be extracted to a workflow step that uses `xpath` or `mvn help:evaluate` like we have in the [test scripts](https://github.com/cryostatio/cryostat/blob/e9ddcaf6503df2e7bc0ef35cc418d0cb01730c7e/run.sh#L10) to just read the `-core` version string and either dump it into the `GITHUB_ENV` or as a step output.

.github/workflows/ci.yaml
```diff
@@ -15,20 +19,141 @@ on:
...
 jobs:
-  build:
+  build-core:
+    runs-on: ubuntu-latest
+    outputs:
+      test_version: ${{ steps.get_core_version.outputs.test_version }}
+    steps:
+    - uses: actions/checkout@v4
+    - id: get_core_version
+      run: echo "test_version=xpath -q -e "project/version/text()" pom.xml" >> $GITHUB_OUTPUT
```
andrewazores, line 29:
```bash
$ echo "test_version=$(xpath -q -e 'project/version/text()' pom.xml)"
test_version=2.5.0-SNAPSHOT
```

mwangggg: thank you!

mwangggg: oh ... no xpath

mwangggg: used mvn help:evaluate instead :smile:

.github/workflows/ci.yaml
```diff
@@ -15,20 +19,134 @@ on:
...
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: 'maven'
+    - run: mvn -B -U clean verify
+    - uses: actions/upload-artifact@v3
+      with:
+        name: cryostat-core
+        path: /home/runner/work/cryostat-core/cryostat-core/target/cryostat-core-${{ steps.get_core_version.outputs.test_version }}.jar
```
andrewazores, line 39: I think `$GITHUB_WORKSPACE` would work as a prefix here:

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

ie. `$GITHUB_WORKSPACE/target/cryostat-core-$version.jar`

You could probably even have `get_core_version` output the `project.build.directory` and use that instead:

```bash
$ ./mvnw -q -DforceStdout -Dexpression=project.build.directory help:evaluate
/home/work/workspace/cryostat-core/target
```

This just makes the workflow a little more generic/reusable.

andrewazores:
```bash
$ ./mvnw -q -DforceStdout -Dexpression=project.build.finalName help:evaluate
cryostat-core-2.25.0-SNAPSHOT
```

.github/workflows/cryostat-build.yml
```diff
@@ -0,0 +1,56 @@
...
+        name: cryostat-core
+    - run: mvn -B -U -Dbuild.arch=${{ inputs.build-arch }} -Dio.cryostat.core.version=${{ inputs.test-version }} clean package
+      env:
+        GITHUB_TOKEN_REF: ${{ secrets.GITHUB_TOKEN }}
+    - name: Save cryostat image
+      run: podman save -o cryostat-${{ inputs.build-arch }}.tar --format oci-archive quay.io/cryostat/cryostat
+    - uses: actions/upload-artifact@v3
+      with:
+        name: cryostat-${{ inputs.build-arch }}
+        path: /home/runner/work/cryostat-core/cryostat-core/cryostat-${{ inputs.build-arch }}.tar
```
andrewazores, line 53: `$GITHUB_WORKSPACE` here too?

.github/workflows/cryostat-IT.yml
tthvo: Hmm, actually from a closer look: 

https://github.com/skjolber/maven-cache-github-action#cache-scopes

tthvo: Looks like the cache is scoped to a specific branch. So PR's workflow runs won't conflict.

.github/workflows/ci.yaml
```diff
@@ -15,20 +19,136 @@ on:
...
+      run: |
+        echo "test_version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)" >> $GITHUB_OUTPUT
+        echo "name=$(mvn help:evaluate -Dexpression=project.artifactId -q -DforceStdout)" >> $GITHUB_OUTPUT
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: 'maven'
+    - run: mvn -B -U clean install
+    - uses: actions/upload-artifact@v3
```
tthvo, line 38: See: https://github.com/skjolber/maven-cache-github-action#cache-scopes

According to this, the cache should not conflict between concurrent workflow (i.e. multiple PRs). I think we can just restore the cache in image build and remove this.

.github/workflows/cryostat-IT.yml
tthvo: Perhaps, its the same pattern for caching. So, `setup-java` should have cache scoped to branches too.

tthvo: image

.github/workflows/ci.yaml
```diff
@@ -15,20 +19,136 @@ on:
...
+      id: arm64_image
+      run: echo "image=${{ steps.push-to-ghcr.outputs.registry-path }}" >> "$GITHUB_OUTPUT"
 
+  cryostat-test:
+    runs-on: ubuntu-latest
+    needs: [push-to-ghcr]
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-java@v2
+    - name: Install xpath
```
tthvo, line 94: Can we use `mvn`? I think the integration script should check for `xpath` and fall back to `mvn`?

.github/workflows/ci.yaml
```diff
@@ -15,20 +19,136 @@ on:
       - cryostat-v[0-9]+.[0-9]+
```
tthvo, line 19: We will need to change to the event to `pull_request_target` to ensure workflow can request write permissions.

tthvo (commented):
See https://github.com/cryostatio/cryostat-core/pull/266#discussion_r1348133672

If its true that cache is scoped to refs, we can avoid conflicts between concurrent workflows: https://github.com/actions/cache#cache-scopes

So, `setup-java` with `cache: maven` should cache and restore the `-core` build in a specific workflow run (i.e. a specific PR). Lets try it out?

.github/workflows/cryostat-build.yml
```diff
@@ -0,0 +1,56 @@
...
+      uses: s4u/maven-settings-action@v2
+      with:
+        servers: '[{"id": "github", "username": "dummy", "password": "${env.GITHUB_TOKEN_REF}"}]'
+        githubServer: false
+    - uses: skjolber/maven-cache-github-action@v1
+      with:
+        step: restore
+    - run: git submodule init
+    - run: git submodule update
+    - uses: actions/download-artifact@v3
```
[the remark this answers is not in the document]
mwangggg: yea maven is separate from this action

mwangggg:
testing :https://github.com/mwangggg/cryostat-core/pull/2

.github/workflows/cryostat-build.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+        repository: ${{ github.repository_owner }}/cryostat
+        ref: main
+        submodules: true
+        fetch-depth: 0
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
```
tthvo, line 31: https://github.com/actions/setup-java#caching-packages-dependencies

Sadly, cache key used hash of "pom.xml" which won't work for our case. Have to use action/cache directly.

Tho, Can we use these configs to set up maven login?

https://github.com/actions/setup-java#maven-options

PR #275, opened 2023-10-16
feat(ci): test -core's build in Cryostat image 

mwangggg:
fixes: #221
(accidentally closed previous PR: #266)


mwangggg:
The integration tests are currently failing, but I'm not sure if it's because the -core version is incompatible or because of something else: https://github.com/mwangggg/cryostat-core/actions/runs/6537266768

andrewazores:
Can you try to trigger that CI run again and see if it fails the same way?

mwangggg:
new CI run: https://github.com/mwangggg/cryostat-core/actions/runs/6539224769

andrewazores:
Okay, that's a very different integration test failure. The first looks like it could have been a cascade of failures after something broke with target discovery during the test. The second looks like https://github.com/cryostatio/cryostat/issues/1661 or something very similar.

mwangggg:
hmm seems like the first run I commented was a rare occassion

mwangggg:
just so I don't forget- Thuan and I were discussing adding the ability to checkout a related PR in Cryostat to test both PRs in Cryostat and -core at the same time.

mwangggg:
I can't believe I spent all this time trying to figure out why the integration tests were failing, only to realize it was because I accidentally ran the tests with the arm64 image instead of the amd64 image. Successful run: https://github.com/mwangggg/cryostat-core/actions/runs/6578935994/job/17873725369

.github/workflows/ci.yaml
```diff
@@ -1,34 +1,161 @@
...
+      run: |
+        echo "core_version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)" >> $GITHUB_OUTPUT
+        echo "name=$(mvn help:evaluate -Dexpression=project.artifactId -q -DforceStdout)" >> $GITHUB_OUTPUT
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+    - uses: actions/cache@v3
+      with:
+        path: ~/.m2
```
aali309, line 45: I think this is what I was missing @andrewazores the other day you mentioned.

mwangggg: oh yea Thuan and I spent a long time trying to figure out how to successfully cache the maven files haha

aali309: Looks good to me.

.github/workflows/ci.yaml
```diff
@@ -1,34 +1,161 @@
...
+        fetch-depth: 0
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
         distribution: 'temurin'
-    - uses: skjolber/maven-cache-github-action@v1
+    - uses: actions/cache@v3
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-build-${{ hashFiles('**/pom.xml') }}
```
andrewazores, line 75: So was the resolution that this `actions/cache` + the `maven-settings-action` need to be used together in order to get both caching and `ghcr.io`/GitHub Maven Packages working, rather than using the `cache:` configuration of `actions/setup-java`?

mwangggg: yes- the reason was because we needed the `restore-keys` option that `actions/cache` gives us since cryostat and -core both have a pom.xml file so the `key` wouldn't produce a hit.

andrewazores: Cache conflicts in a GitHub Actions CI pipeline. Fun.

andrewazores (approved):

.github/workflows/ci.yaml
mwangggg: also, I tried to use the server setup options in `actions/setup-java`, but I don't think it was adding it to the right place? `maven-settings-action` adds the cryostat server to `settings.xml` so we kept it the way it is.

Issue #277, opened 2023-10-19
[Task] CI on push to `main`


andrewazores:
In https://github.com/cryostatio/cryostat-core/pull/275 we got some nice new PR CI features, but we lost the CI workflow that runs on push to `main`, ie. when a PR is merged. It should just a be a matter of running through the old pre-275 workflow again, basically `mvn verify`, so that we can have that extra assurance that `main` is really in a good buildable state.

PR #279, opened 2023-10-19
chore(ci): restore CI on push workflow

mwangggg:
fixes: #277

### DIFF
--- a/.github/workflows/push-ci.yml
+++ b/.github/workflows/push-ci.yml
@@ -1,0 +1,20 @@
+name: CI build and push
+
+on:
+  push:
+    branches:
+      - main
+      - v[0-9]+
+      - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+    - run: mvn -B -U clean verify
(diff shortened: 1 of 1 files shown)


mwangggg:
not sure if I should update `setup-java` version to 21?

andrewazores:
I was actually about to say the opposite. The published artifact builds with JDK 11 for wider compatibility: https://github.com/cryostatio/cryostat-core/blob/17a29a3daff78f69de36864817bd504a5c7b5473/.github/workflows/maven-publish.yaml#L26

so I think this should either match and build with 11 only, or it should be an 11/17/21 matrix.

mwangggg:
I'll leave it as 11 then

andrewazores:
I, uh, think that was a little excessive :sweat_smile: 

I suspect this is the problem: https://github.com/cryostatio/cryostat-core/blob/17a29a3daff78f69de36864817bd504a5c7b5473/.github/workflows/ci.yaml#L4

The concurrency group is based on the run ID, but each time the PR is relabelled or a new commit is pushed I think that triggers a new run with a new ID, so many concurrent jobs can be running in the background on the same actual code if we're just relabelling the PR etc.

andrewazores:
Just to test that out ^

https://github.com/cryostatio/cryostat-core/actions/runs/6580357398

andrewazores:
Nice, it passed - would be even nicer if the status got reported, either as a comment again like we already have elsewhere, or [as a status check item](https://josh-ops.com/posts/using-github-checks-api/)

andrewazores (approved):

Issue #281, opened 2023-10-20
[Bug] Excessive/duplicate CI runs


andrewazores:
I, uh, think that was a little excessive :sweat_smile: 

I suspect this is the problem: https://github.com/cryostatio/cryostat-core/blob/17a29a3daff78f69de36864817bd504a5c7b5473/.github/workflows/ci.yaml#L4

The concurrency group is based on the run ID, but each time the PR is relabelled or a new commit is pushed I think that triggers a new run with a new ID, so many concurrent jobs can be running in the background on the same actual code if we're just relabelling the PR etc.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-core/issues/279#issuecomment-1771718706_

PR #290, opened 2023-10-31
feat(ci): add status check for new workflow

aali309:
fixes: #283

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,7 +54,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Install qemu
-      if: ${{ inputs.build-arch != 'amd64' }}
+      if: ${{ matrix.arch != 'amd64' }}
       continue-on-error: true
       run: |
         sudo apt-get update
@@ -98,7 +98,7 @@ jobs:
     strategy:
       matrix:
         arch: [amd64, arm64]
-    outputs: 
+    outputs:
       amd64_image: ${{ steps.amd64_image.outputs.image }}
       arm64_image: ${{ steps.arm64_image.outputs.image }}
     permissions:
@@ -141,7 +141,7 @@ jobs:
     - name: Create markdown table
       id: md-table
       uses: petems/csv-to-md-table-action@v3.0.0
-      with: 
+      with:
         csvinput: |
           ARCH, IMAGE
           amd64, ${{ env.amd64_image }}
@@ -153,7 +153,7 @@ jobs:
 
           To run smoketest:
           ```
-          # amd64          
+          # amd64
           CRYOSTAT_IMAGE=${{ env.amd64_image }} sh smoketest.sh
 
           # or arm64
--- a/.github/workflows/cryostat-test.yml
+++ b/.github/workflows/cryostat-test.yml
@@ -41,10 +41,10 @@ jobs:
               content: "+1",
             });
 
-  checkout-branch: 
+  checkout-branch:
     runs-on: ubuntu-latest
     needs: [check-before-build]
-    outputs: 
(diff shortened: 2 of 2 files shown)


aali309:
pr [LINK]: (https://github.com/aali309/cryostat-core/pull/2)

image

image

aali309:
@mwangggg its complaining about pulling the image when I test it locally?

mwangggg:
ah it failed because an image hasn't been built for this specific ref for your PR.. perhaps the `CI build` workflow needs to be triggered on more PR events- I'll take a look

mwangggg:
oh it's because you ran `/build_test` before the `CI build` workflow was completed- so those images hadn't been pushed to ghcr.io yet

andrewazores:
I guess putting the workflows in the same concurrency group and using `cancel-in-progress: false` could be a useful queuing behaviour for that kind of case, but I don't know if there would be other downsides associated with that, and I haven't put any thought into what to use as the concurrency group ID to tie those things together - I guess it would just have to be the branch name/checkout ref?

aali309:
> oh it's because you ran `/build_test` before the `CI build` workflow was completed- so those images hadn't been pushed to ghcr.io yet

yes, I figured after looking at the build again and I deleted the comment

mwangggg:
Hmm if they're in the same concurrency group and the `/build_test` workflow is running, I guess whomever's PR it is will have to wait until it's done to push any changes - could cause an old image to be tested if they're not careful. But otherwise, I can't think of any other downsides.

aali309:
> @andrewazores Hmm if they're in the same concurrency group and the `/build_test` workflow is running, I guess whomever's PR it is will have to wait until it's done to push any changes - could cause an old image to be tested if they're not careful. But otherwise, I can't think of any other downsides?

This is what I ran into at first.

andrewazores:
Maybe there would be a way to group those things into a concurrency group based on the latest commit hash, so that when a new push comes in that has a new hash and is therefore in a new concurrency group/separate queue? I'm not sure if this is really worth the time investment to get working, it's going to come up pretty rarely and the workaround is just to wait for the initial build completion before invoking the build_test workflow.

aali309:
> Maybe there would be a way to group those things into a concurrency group based on the latest commit hash, so that when a new push comes in that has a new hash and is therefore in a new concurrency group/separate queue? I'm not sure if this is really worth the time investment to get working, it's going to come up pretty rarely and the workaround is just to wait for the initial build completion before invoking the build_test workflow.

I agree about the time it might take to get this to work. May be something we can look at in the future?

mwangggg:
we could add a step that checks if the image exists, and if not, exit with a message that informs the commenter to wait until the initial build and push is completed to try integration testing?

aali309:
> we could add a step that checks if the image exists, and if not, exit with a message that informs the commenter to wait until the initial build and push is completed to try integration testing?

yes, this was what I was thinking as well.

mwangggg:
> I guess putting the workflows in the same concurrency group and using `cancel-in-progress: false` could be a useful queuing behaviour for that kind of case, but I don't know if there would be other downsides associated with that, and I haven't put any thought into what to use as the concurrency group ID to tie those things together - I guess it would just have to be the branch name/checkout ref?

Also, I'm not sure if it's the best idea to have `cancel-in-progress: false` for these workflows. There could be lots of bot comments from relabelling etc. i.e.: https://github.com/cryostatio/cryostat-core/issues/281

aali309:
Took a lot of time, to conclude with this one. Many other solutions available but they will require a lot of code change. May be later we can look to make it like `cryostat main` to adopt the  `retest` option?
image

andrewazores:
Looks pretty good to me. Rebase please

Issue #295, opened 2023-11-14
Spotless fails build after update to 4.8.1


aptmac:
Spotbugs 4.8.0 introduced a new detector for `CT_CONSTRUCTOR_THROW` [[0]](https://github.com/spotbugs/spotbugs/blob/4.8.0/CHANGELOG.md#added), which flags constructors that throw exceptions and may be left in a partially initialized and accessible state, see: [OBJ11-J [1]](https://wiki.sei.cmu.edu/confluence/display/java/OBJ11-J.+Be+wary+of+letting+constructors+throw+exceptions).

See: https://github.com/cryostatio/cryostat-core/actions/runs/6786037630/job/18445631568#step:4:1634
```
[INFO] --- spotbugs-maven-plugin:4.8.1.0:check (spotbugs) @ cryostat-core ---
[INFO] BugInstance size is 7
[INFO] Error size is 0
[INFO] Total bugs: 7
[ERROR] Medium: Exception thrown in class io.cryostat.core.EventOptionsBuilder at new io.cryostat.core.EventOptionsBuilder(ClientWriter, JFRConnection, Supplier) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks. [io.cryostat.core.EventOptionsBuilder, io.cryostat.core.EventOptionsBuilder] At EventOptionsBuilder.java:[line 50]At EventOptionsBuilder.java:[line 50] CT_CONSTRUCTOR_THROW
[ERROR] Medium: Exception thrown in class io.cryostat.core.agent.AgentJMXHelper at new io.cryostat.core.agent.AgentJMXHelper(IConnectionHandle) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks. [io.cryostat.core.agent.AgentJMXHelper, io.cryostat.core.agent.AgentJMXHelper] At AgentJMXHelper.java:[line 45]At AgentJMXHelper.java:[line 45] CT_CONSTRUCTOR_THROW
[...]

PR #296, opened 2023-11-14
build(deps): downgrade spotbugs

andrewazores:
See #295 #294 #272

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
   <org.jsoup.version>1.16.1</org.jsoup.version>
 
   <com.github.spotbugs.version>4.8.0</com.github.spotbugs.version>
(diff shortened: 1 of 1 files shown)


andrewazores:
What's happening with the CI here... ? Is it not checking out the PR but `main` instead? I didn't notice the Spotbugs failures on the PR upgrading the version, and here where I am downgrading the version it is still running the newer failing version. @aali309 @mwangggg

aptmac (approved):

aali309:
> What's happening with the CI here... ? Is it not checking out the PR but `main` instead? I didn't notice the Spotbugs failures on the PR upgrading the version, and here where I am downgrading the version it is still running the newer failing version. @aali309 @mwangggg

Was this resolved?

andrewazores:
No, I just ignored it and merged the Spotbugs downgrade. The CI still appeared to be picking up the newer version from `main` and failing the PR CI with it.

Issue #315, opened 2023-12-04
[Bug] CI is not correctly caching the PR's -core version


mwangggg:
```
 [INFO] Scanning for projects...
[INFO] 
[INFO] ------------------------< io.cryostat:cryostat >------------------------
[INFO] Building cryostat 2.5.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] Downloading from github: https://maven.pkg.github.com/cryostatio/cryostat-core/io/cryostat/cryostat-core/2.26.0-SNAPSHOT/maven-metadata.xml
[INFO] Downloading from github: https://maven.pkg.github.com/cryostatio/cryostat-core/io/cryostat/cryostat-core/2.26.0-SNAPSHOT/cryostat-core-2.26.0-SNAPSHOT.pom
Warning:  The POM for io.cryostat:cryostat-core:jar:2.26.0-SNAPSHOT is missing, no dependency information available
[INFO] Downloading from github: https://maven.pkg.github.com/cryostatio/cryostat-core/io/cryostat/cryostat-core/2.26.0-SNAPSHOT/cryostat-core-2.26.0-SNAPSHOT.jar
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.642 s
[INFO] Finished at: 2023-12-04T18:45:26Z
[INFO] ------------------------------------------------------------------------
Error:  Failed to execute goal on project cryostat: Could not resolve dependencies for project io.cryostat:cryostat:jar:2.5.0-SNAPSHOT: Could not find artifact io.cryostat:cryostat-core:jar:2.26.0-SNAPSHOT in github (https://maven.pkg.github.com/cryostatio/cryostat-core) -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
Error: Process completed with exit code 1.
```

CI failure looks like the builder is trying to use `2.26.0-SNAPSHOT` for `-core` because that's the version in the `pom.xml` in this PR (since that's what is on `main`), but that is not a published version since we only publish actual releases, not snapshots.

[...]

PR #317, opened 2023-12-05
chore(ci): checkout core first to cache

mwangggg:
fixes: #315
Basically, when caching the maven files, the key that we use is unique to each version of the `pom.xml` file. When we checkout the `Cryostat` repo, which has a different `pom.xml` file, we are unable to hit the exact key that we saved the cache under because the hashes of the two different `pom.xml` files are different. Additionally, we are not hitting the cache that was most recently produced for some reason. Therefore, sometimes with changes like the update to `2.26.0-SNAPSHOT`, we are actually restoring an older version of -core that hasn't been updated yet.
Here, the change I've made just checks out the -core PR branch before checking out the main `Cryostat` branch so the right files are cached into the `~/.m2` folder.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,18 @@ jobs:
         sudo apt-get install -y qemu-user-static
     - uses: actions/checkout@v4
       with:
+        repository: ${{ github.event.pull_request.head.repo.full_name }}
+        ref: ${{ github.event.pull_request.head.ref }}
+        fetch-depth: 0
+    - uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


mwangggg:
demo: https://github.com/mwangggg/cryostat-core/actions/runs/7102397131/job/19332668379

andrewazores:
That failure is the same as this: https://github.com/cryostatio/cryostat/pull/1808

Cryostat (and Cryostat3) need a small update to be buildable with the latest `-core` release.

andrewazores (approved):

Issue #39, opened 2020-06-23
Log format should be configurable


andrewazores:
Log format should be configurable to rearrange fields, and include extra fields not currently supported (ex. timestamp). This may mean removing the actual logger implementation and wrapping the abstraction around another lightweight logging library, if need be.

PR #322, opened 2023-12-18
feat(logs): configure log format

aali309:
fixes: #39

### DIFF
--- a/src/main/java/io/cryostat/core/log/LogConfig.java
+++ b/src/main/java/io/cryostat/core/log/LogConfig.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright The Cryostat Authors.
+ *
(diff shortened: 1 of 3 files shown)


andrewazores:
I'm not sure this is quite the right approach - rather than the configuration only allowing for booleans of whether information should be included in the output or not, it would be more flexible and more conventional to use formatting specifiers, so the configuration should be set as ex. an environment variable that looks like `LOG_FMT=[%t] (%l) %m`, where `%t` would be replaced by the timestamp, `%l` by the log level, `%m` by the log message, for example.

Regardless, this is a pretty low priority request - it has been open for 3.5 years and hasn't really been a real problem, so I would rather we focus our efforts on more immediately fruitful development :-)

Issue #349, opened 2024-02-14
[Bug] CI failing to build Cryostat3


mwangggg:
https://github.com/cryostatio/cryostat-core/actions/runs/7905086789/job/21576921923?pr=348#step:13:29

The CI can't retrieve the latest version of Cryostat core

PR #350, opened 2024-02-14
fix(CI): checkout -core to cache mvn files before building Cryostat3

mwangggg:
fixes: #349

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,11 @@ jobs:
         pull-requests: write
         statuses: write
     steps:
+    - uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -122,6 +122,11 @@ jobs:
         pull-requests: write
         statuses: write
     steps:
+    - uses: actions/checkout@v4
+      with:
+        repository: ${{ github.event.pull_request.head.repo.full_name }}
+        ref: ${{ github.event.pull_request.head.ref }}
+        fetch-depth: 0
```
andrewazores, line 129: Is checking it out enough or does it also need to do a `mvn install` to get the latest version into `~/.m2`?

PR #348, opened 2024-02-12
feat(logs): remove logger abstraction

mwangggg:
fixes: #52

PR #354, opened 2024-02-22
chore(logger): do not pass logger instances into constructors

andrewazores:
Related to #348

No need to pass in logger instances anymore, they can and should just use the LoggerFactory to get an instance for themselves.


aali309 (approved):
looks good to me.

PR #423, opened 2024-07-18
chore(multimodule): split out libcryostat module

### LINKED ISSUE #416: Split implementations from interfaces and simple models
Most of the Cryostat project is happily using JDK 17 and will upgrade to JDK 21 when other dependencies upgrade. However, the Cryostat Agent will always want to lag behind this and remain compatible with a wider range of JDK versions. At the moment this means supporting JDK 11 as well. However, the JMC dependencies (#173) used here in -core now require JDK17+.

We should split this library up into parts that are JDK17+ or JDK21+, including all of the JMC-related pieces, and a part that is JDK-11 compatible including the few small pieces that are used by the Agent. This split off library could be called something like `cryostat-interfaces` or `cryostat-models`, and would contain only things like interface definitions or simple classes used for serialization. Then the Agent would only use this new dependency, and the other containers/components can use -core and upgrade freely to newer JDK versions.

andrewazores:
Fixes #416


andrewazores:
CI failure makes sense - quite a few classes got moved from `cryostat-core` into `libcryostat`. Most are directly moved with no functional changes but a few had some other minor changes, so all of the previous `cryostat-core` consumers are likely to need adjustments to import classes from `libcryostat` to compile again.

ebaron (approved):

aptmac (approved):
LGTM.

On another (but kind of related) note, it looks like rjmx.common and flightrecorder.configuration got uploaded to Maven Central, so the Adoptium repository can be removed from the poms. https://central.sonatype.com/namespace/org.openjdk.jmc

PR #430, opened 2024-07-22
ci(modules): build modules using maven toolchains

### LINKED ISSUE #425: [CI] `libcryostat` should not only target JDK 11, but should be built with JDK 11

andrewazores:
- **build(toolchain): set up toolchains plugin, configure poms**
- **ci(toolchains): install JDKs 11 and 17**
- **ci(cryostat): build Cryostat with single JDK version only**

See #426
Fixes #425

Test run: https://github.com/andrewazores/cryostat-core/actions/runs/10043497824/job/27756411088

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,9 @@ jobs:
         echo "name=$(mvn help:evaluate -Dexpression=project.artifactId -q -DforceStdout)" >> $GITHUB_OUTPUT
     - uses: actions/setup-java@v3
       with:
-        java-version: '17'
+        java-version: |
+          11
+          17
         distribution: 'temurin'
     - uses: actions/cache@v3
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,12 +59,9 @@ jobs:
   build-cryostat:
     needs: [build-core]
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        java: ['17']
     env:
         cache-name: cache-yarn
-    name: Build Java ${{ matrix.java }}
+    name: Build Cryostat
     permissions:
         packages: write
         contents: read
@@ -91,7 +88,7 @@ jobs:
         fetch-depth: 0
     - uses: actions/setup-java@v3
       with:
-        java-version: ${{ matrix.java }}
+        java-version: '17'
         distribution: 'temurin'
         cache: 'maven'
     - name: maven-settings
@@ -132,7 +129,6 @@ jobs:
       run: echo "DOCKER_HOST=unix:///run/user/$(id -u)/podman/podman.sock" >> "$GITHUB_ENV"
     - name: Build cryostat
       run: ./mvnw -B -U -Dio.cryostat.core.version=${{ needs.build-core.outputs.core_version }} clean verify
-      continue-on-error: ${{ matrix.java != '17' }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Debug podman images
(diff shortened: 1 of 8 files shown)


mwangggg (approved):
lgtm

andrewazores:
I *think* this worked. https://s01.oss.sonatype.org/content/repositories/snapshots/io/cryostat/

```bash
d=$(mktemp -d)
pushd $d
mvn -U dependency:copy -DoutputDirectory=. -Dartifact=io.cryostat:cryostat-core:4.0.0-SNAPSHOT
unzip cryostat-core-4.0.0-SNAPSHOT.jar -d cryostat-core
pushd cryostat-core
javap -verbose io/cryostat/core/CryostatCore.class | grep "major version" # indicates 61, ie JDK 17
cat META-INF/MANIFEST.MF | grep Build # indicates JDK 17
popd
mvn -U dependency:copy -DoutputDirectory=. -Dartifact=io.cryostat:cryostat-core:4.0.0-SNAPSHOT
unzip libcryostat-4.0.0-SNAPSHOT.jar -d cryostat-core
pushd libcryostat
javap -verbose io/cryostat/libcryostat/JvmIdentifier.class | grep "major version" # indicates 55, ie JDK 11
cat META-INF/MANIFEST.MF | grep Build # indicates JDK 17 as well?
popd
popd
```

andrewazores:
I'm guessing the manifests are wrong because the overall packaging was done by JReleaser using JDK 17, since that's the default JDK for the entire project. But since libcryostat builds using JDK 11 specifically, the classes contained in its JAR do indicate the correct bytecode version, even though the manifest suggests something else.

